PR #37, opened 2019-08-10
chore: prepare to implement Gameobject Editor

Helias:
Implemented types and services


src/app/services/editors/gameobject/gameobject-loot-template.service.ts
```diff
@@ -0,0 +1,35 @@
...
+import { LootEditorService } from '../loot-editor.service';
+import {
+  GAMEOBJECT_TEMPLATE_ID, GAMEOBJECT_TEMPLATE_LOOT_ID,
+  GAMEOBJECT_TEMPLATE_TABLE
+} from '../../../types/gameobject-template.type';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class GameobjectLootTemplateService extends LootEditorService<GameobjectLootTemplate> {
```
FrancescoBorzi: the extra logic added in `LootEditorService` is only handling the part when a loot ID must be assigned in the main template table of the entity. For example `creature_loot_template` is linked to `lootId` of `creature_template`, skinning/pickpocketing loot also have their linked IDs.

Does `gameobject_loot_template` have a loot id in `gameobject_template` ? If not, it should just extend `MultiRowEditorService` instead

src/app/types/gameobject-spawn.type.ts
```diff
@@ -0,0 +1,28 @@
+import { TableRow } from './general';
+
+export const SPAWNS_TABLE = 'gameobject';
+export const SPAWNS_ID = 'id';
+export const SPAWNS_ID_2 = 'guid';
```
FrancescoBorzi: those contants should be renamed in `GAMEOBJECT_SPAWNS_*`

src/app/types/gameobject-template.type.ts
```diff
@@ -0,0 +1,50 @@
+import { TableRow } from './general';
+
+export const GAMEOBJECT_TEMPLATE_TABLE = 'gameobject_template';
+export const GAMEOBJECT_TEMPLATE_ID = 'entry';
+export const GAMEOBJECT_TEMPLATE_NAME = 'name';
+export const GAMEOBJECT_TEMPLATE_CUSTOM_STARTING_ID = 9_000_000;
```
FrancescoBorzi: the current highest **entry** value in `gameobject_template` is `400_000` so we could make the custom start from `900_000` (one less zero)

FrancescoBorzi:
@Helias also the build is failing:

https://travis-ci.com/azerothcore/Keira3/builds/122699152

src/app/types/gameobject-template.type.ts
```diff
@@ -0,0 +1,50 @@
...
+  Data15: number = 0;
+  Data16: number = 0;
+  Data17: number = 0;
+  Data18: number = 0;
+  Data19: number = 0;
+  Data20: number = 0;
+  Data21: number = 0;
+  Data22: number = 0;
+  Data23: number = 0;
+  AIName: char   = '';
```
FrancescoBorzi: should be `string`

src/app/services/editors/gameobject/gameobject-loot-template.service.ts
Helias: I'll solve this later.
I will consider this:
```
Only gameobject type 3 (GAMEOBJECT_TYPE_CHEST) or 25 (GAMEOBJECT_TYPE_FISHINGHOLE) use data1 as loot ID, for other types data1 is used in other ways
```
[[source]](http://www.azerothcore.org/wiki/loot_template#loot_template-Entry)

PR #55, opened 2019-08-18
Added Gameobject Template Addon, added minor fixes

Helias:
Added Gameobject Template Addon.

Fixed gamebject_questitem and gameobject_template_addon types (table name).

### DIFF
--- a/src/app/components/editors/gameobject/gameobject-template-addon/gameobject-template-addon.integration.spec.ts
+++ b/src/app/components/editors/gameobject/gameobject-template-addon/gameobject-template-addon.integration.spec.ts
@@ -92,7 +92,8 @@ describe('GameobjectTemplateAddon integration tests', () => {
     });
 
     it('changing all properties and executing the query should correctly work', () => {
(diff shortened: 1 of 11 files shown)


src/app/components/editors/gameobject/gameobject-template-addon/gameobject-template-addon.integration.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+    });
+
+    it('changing values should correctly update the queries', () => {
+      page.setInputValueById('faction', '35');
+      page.expectDiffQueryToContain(
+        'UPDATE `gameobject_template_addon` SET `faction` = 35 WHERE (`entry` = ' + id + ');'
+      );
+      page.expectFullQueryToContain('35');
+
+    });
```
FrancescoBorzi: add one test case that uses the FlagsSelector to change the value, you can look at this example:

https://github.com/azerothcore/Keira3/blob/master/src/app/components/editors/quest/quest-template-addon/quest-template-addon.integration.spec.ts#L165

PR #57, opened 2019-08-18
Added Gameobject Questitem

Helias:
Added Gameobject Questitem, fixed types and added minor improvements


src/app/components/main-window/sidebar/sidebar.component.html
```diff
@@ -214,6 +214,11 @@
                       Template Addon
                     </a>
                   </li>
+                  <li>
+                    <a href="#" routerLinkActive="active" [routerLink]="'gameobject/gameobject-questitem'">
+                      Quest Items
+                    </a>
+                  </li>
```
FrancescoBorzi: all editors except the main one (in this case `gameobject_template`) should be moved in the `*ngIf="!gameobjectHandlerService.isNew"` below

PR #517, opened 2020-03-23
feat: display item names

FrancescoBorzi:
- Closes https://github.com/azerothcore/Keira3/issues/512

### DIFF
--- a/src/app/shared/services/mysql-query.service.spec.ts
+++ b/src/app/shared/services/mysql-query.service.spec.ts
@@ -803,60 +803,84 @@ describe('MysqlQueryService', () => {
   });
 
   describe('get helpers', () => {
-    const result = of('mock result');
-    const resultToPromise = result.toPromise();
+    const result = 'mock result';
+    const resultToObs = of(result);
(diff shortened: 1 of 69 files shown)


src/app/shared/services/mysql-query.service.spec.ts
```diff
@@ -803,60 +803,84 @@ describe('MysqlQueryService', () => {
...
+      spyOn(service, 'queryValue').and.returnValue(resultToObs);
       spyOn(service, 'queryValueToPromise').and.returnValue(resultToPromise);
     });
 
-    it('getCreatureNameById', () => {
-      expect(service.getCreatureNameById(id)).toEqual(resultToPromise);
-      expect(service.queryValueToPromise).toHaveBeenCalledWith(
+    it('clearCache', async () => {
+      expect(await service.getCreatureNameById(id)).toEqual(result);
+      expect(await service.getCreatureNameById(id)).toEqual(result);
```
Helias: This code is redundant, we should move it in a function or use beforeEach()

PR #550, opened 2020-03-29
feat(sidebar): show unsaved changes

FrancescoBorzi:
- Closes https://github.com/azerothcore/Keira3/issues/549

### DIFF
--- a/src/app/shared/abstract/service/handlers/handler.service.ts
+++ b/src/app/shared/abstract/service/handlers/handler.service.ts
@@ -8,10 +8,8 @@ export abstract class HandlerService<T extends TableRow> extends SubscriptionHan
   selectedName: string;
   isNew = false;
 
-  protected abstract _statusMap: {
-    [key: string]: boolean;
-  };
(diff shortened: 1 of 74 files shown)


src/app/shared/abstract/service/handlers/handler.service.ts
```diff
@@ -21,7 +28,19 @@ export abstract class HandlerService<T extends TableRow> extends SubscriptionHan
     super();
   }
 
+  private resetStatus() {
```
Helias: add the return type of the function

src/app/shared/abstract/service/handlers/handler.service.ts
```diff
@@ -8,6 +8,13 @@ export abstract class HandlerService<T extends TableRow> extends SubscriptionHan
   selectedName: string;
   isNew = false;
 
+  protected abstract _statusMap: {
+    [key: string]: boolean;
+  };
+  get statusMap() {
```
Helias: add the return type

PR #567, opened 2020-04-03
test(item-preview): add integration test

Helias:


src/app/features/item/item-template/item-preview.service.ts
```diff
@@ -555,7 +555,7 @@ export class ItemPreviewService {
     let dmg = '';
 
     if (itemClass === ITEM_TYPE.AMMUNITION && dmgmin && dmgmax) {
-      if (sc1) {
+      if (sc1 && sc1 < 7 && sc1 > 0) {
```
FrancescoBorzi, line 558: wrap `7` into a constants

PR #595, opened 2020-04-09
chore(creature-spawn.type): rename 'spawndist' to 'wander_distance'

Kitzunu:
This will be necessary when https://github.com/azerothcore/azerothcore-wotlk/pull/2858 is pushed. 

This is the only file I found with spawndist in it... if there is another just poke me :P


FrancescoBorzi:
Tested.

Before the PR:

image

With the PR:

image

good job @Kitzunu !

PR #661, opened 2020-05-03
chore(quest-preview): show provided item

FrancescoBorzi:


src/app/features/quest/quest-preview/quest-preview.service.spec.ts
```diff
@@ -117,6 +117,19 @@ describe('QuestPreviewService', () => {
     expect(service.sharable).toBe('Not sharable');
   });
 
+  it('start item', async () => {
+    const { service, questTemplateService, mysqlQueryService } = setup();
+    const mockStartItem = 123456;
+    const mockStartItemName = 'Sword of AzerothCore';
+    spyOn(mysqlQueryService, 'getItemNameById').and.callFake(() => of(mockStartItemName).toPromise());
```
Helias, line 124: use `Promise.resolve()` instead of `of().toPromise()` like above

PR #706, opened 2020-05-22
feat(conditions): add quest-state flag selector

FrancescoBorzi:


src/app/shared/constants/flags/quest-state.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export const QUEST_STATE: Flag[] = [
+  { bit: 0, name: 'NONE (not taken)', },
+  { bit: 1, name: 'COMPLETE', },
+  { bit: 2, name: 'UNAVAILABLE (unused)', },
+  { bit: 3, name: 'INCOMPLETE (in progress)', },
+  { bit: 4, name: 'AVAILABLE (unused)', },
+  { bit: 5, name: 'FAILED', },
+  { bit: 6, name: 'REWARDED', },
+];
```
FrancescoBorzi, line 11: @Kitzunu  1) this creates the Flag definition. The `bit` value are powers of 2.

src/app/features/conditions/edit-conditions/conditions.component.ts
```diff
@@ -36,6 +37,8 @@ export class ConditionsComponent extends SingleRowEditorComponent<Conditions> {
   public readonly CONDITION_VALUE_2_TOOLTIPS = CONDITION_VALUE_2_TOOLTIPS;
   public readonly CONDITION_VALUE_3_TOOLTIPS = CONDITION_VALUE_3_TOOLTIPS;
 
+  public readonly QUEST_STATE = QUEST_STATE;
```
FrancescoBorzi, line 40: @Kitzunu  2) here we simply import the flag in order to make it available for the component's view (HTML)

src/app/features/conditions/edit-conditions/conditions.component.ts
```diff
@@ -44,6 +47,10 @@ export class ConditionsComponent extends SingleRowEditorComponent<Conditions> {
     return this.editorService.form.controls.ConditionTypeOrReference.value;
   }
 
+  get showQuestState(): boolean {
+    return this.conditionType === CONDITION_TYPES.CONDITION_QUESTSTATE;
+  }
```
FrancescoBorzi, line 52: @Kitzunu  3) small method that is telling us when to show the quest state flag selector (according to the value of condition type)

src/app/features/conditions/edit-conditions/conditions.component.html
```diff
@@ -103,6 +103,12 @@
               class="fas fa-info-circle ml-1" [placement]="'auto'"
               [tooltip]="CONDITION_VALUE_2_TOOLTIPS[conditionType]"
             ></i>
+            <keira-flags-selector-btn
+              *ngIf="showQuestState"
+              [control]="editorService.form.controls.ConditionValue2"
+              [config]="{ flags: QUEST_STATE, name: 'ConditionValue2' }"
+              [modalClass]="'modal-lg'"
+            ></keira-flags-selector-btn>
```
FrancescoBorzi, line 111: @Kitzunu  4) and, finally, this is the selector added to the view

src/app/shared/constants/flags/quest-state.ts
```diff
@@ -0,0 +1,11 @@
+import { Flag } from '../../types/general';
+
+export const QUEST_STATE: Flag[] = [
+  { bit: 0, name: 'NONE (not taken)', },
+  { bit: 1, name: 'COMPLETE', },
+  { bit: 2, name: 'UNAVAILABLE (unused)', },
+  { bit: 3, name: 'INCOMPLETE (in progress)', },
+  { bit: 4, name: 'AVAILABLE (unused)', },
```
Kitzunu, line 8: Why are they marked as unused?

FrancescoBorzi: @Kitzunu

because they seem to be unused in the core:

image

Kitzunu: Actually this condition doesnt even make sense because we have 
CONDITION_QUESTREWARDED
CONDITION_QUESTTAKEN
CONDITION_QUEST_NONE
CONDITION_QUEST_COMPLETE

Only one that doesn't exist is Failed
Perhaps this condition is just so we can have less queries if we want to make conditions with all of them. But then, why do they exist when we can just use this?
weird

PR #716, opened 2020-05-24
feat(conditions): add type mask flag selector

Kitzunu:


Kitzunu:
To explain the flags:

According to the wiki these are the only ones that are in use and therefore I don't want to add the other flags to prevent confusion for the end user

PR #717, opened 2020-05-24
feat(Conditions/Options): Implement option select for conditions

Kitzunu:


src/app/features/conditions/edit-conditions/conditions.component.ts
```diff
@@ -63,6 +88,42 @@ export class ConditionsComponent extends SingleRowEditorComponent<Conditions> {
...
+    return this.conditionType === CONDITION_TYPES.CONDITION_HP_VAL;
+  }
+
+  get ShowHPPct(): boolean {
+    return this.conditionType === CONDITION_TYPES.CONDITION_HP_PCT;
+  }
+
+  get ShowStandState(): boolean {
+    return this.conditionType === CONDITION_TYPES.CONDITION_STAND_STATE;
+  }
```
FrancescoBorzi, line 125: @Kitzunu method names should be lowercase. E.g. `ShowStandState` -> `showStandState`

Kitzunu: Ahh yes that is very true. Like I did before this commit https://github.com/azerothcore/Keira3/pull/717/commits/4c1a8f1f230e3a6b3c126f211f0caa837ad4b098

I knew I made something wrong :D

FrancescoBorzi:
@Kitzunu 
```
error NG8002: Can't bind to 'control' since it isn't a known property of 'keira-single-value-selector-btn'.
```

to fix this, you need to import the `SingleValueSelectorModule` inside `ConditionsEditorModule`

PR #720, opened 2020-05-24
fix some codacy stuff

Kitzunu:


src/app/main/connection-window/connection-window.component.scss
```diff
@@ -160,7 +164,9 @@ $secondary-color-green: #b7e2ad;
       bottom: 10px;
       left: 50%;
       transform: translateX(-50%);
-      a, a:hover {
+
+      a::before,
+      a:hover::before {
```
Kitzunu, line 169: I dont know if this is correct, that is up to you experts to decide.
But it did fix codacy :D

FrancescoBorzi:
@Kitzunu did you test this? I mean, by having a look around the app and check that all the UI is fine

Kitzunu:
Very true, I totally forgot about this after I fixed the ability to run npm

gonna check it asap

src/app/main/connection-window/connection-window.component.scss
FrancescoBorzi: shouldn't it be:

```
a,
a:hover {
```
I don't see why we should add `::before`

Kitzunu: a,
a:hover 

makes Codacy mad

I read some and ::before fixed the issue, but as I said above I don't know if it is the correct way of doing it

src/assets/scss/_ngx-datatable.scss
```diff
@@ -2,42 +2,50 @@
...
+
       &:hover {
         background-color: #eee;
       }
-      &.active {
+      // &.active {
         //background-color: blue;
         //font-weight: 600;
-      }
+      // }
```
FrancescoBorzi, line 47: ?

Kitzunu: it was complaining that it was "illegal" to have it empty:
```
&.active { }
```
 since the stuff inside it is commented

FrancescoBorzi:
mostly it looks good, just a couple of comments

but please test it well before merging

Kitzunu:
Everything looks good as far as I can see

src/assets/scss/_ngx-datatable.scss
FrancescoBorzi: ah ok ok

src/app/main/connection-window/connection-window.component.scss
FrancescoBorzi: ok I need to have a look at that, everything else is good

PR #929, opened 2020-10-13
Upgrade of older versions of Ajv, Minimist and node-addon-api

MKokeshi:
### Ajv
Ajv is Another JSON Schema Validator and old versions  of this package are vulnerable to Prototype Pollution. A carefully crafted JSON schema could be provided that allows execution of other code by prototype pollution. (While untrusted schemas are recommended against, the worst case of an untrusted schema should be a denial of service, not execution of code.).
**Prototype Pollution** refers to the ability to inject properties into existing JavaScript language construct prototypes, such as objects. JavaScript allows all Object attributes to be altered, including their magical attributes such as `_proto_`,` constructor` and `prototype`. An attacker manipulates these attributes to overwrite, or pollute, a JavaScript application object prototype of the base object by injecting other values. Properties on the Object.prototype are then inherited by all the JavaScript objects through the prototype chain. When that happens, this leads to either denial of service by triggering JavaScript exceptions, or it tampers with the application source code to force the code path that the attacker injects, thereby leading to remote code execution.
The main ways in which the pollution of prototypes occurs:

1. Unsafe Object recursive merge
2. Property definition by path
### Poc of Unsafe Object recursive merge
```
merge (target, source)

  foreach property of source

    if property exists and is an object on both the target and the source

      merge(target[property], source[property])

    else

      target[property] = source[property]
```
When the source object contains a property named `_proto_` defined with `Object.defineProperty()` , the condition that checks if the property exists and is an object on both the target and the source passes and the merge recurses with the target, being the prototype of `Object` and the source of `Object` as defined by the attacker. Properties are then copied on the `Object` prototype.

Clone operations are a special sub-class of unsafe recursive merges, which occur when a recursive merge is conducted on an empty object: `merge({},source).
`
`lodash` and `Hoek` are examples of libraries susceptible to recursive merge attacks.
**Property definition by path**
There are a few JavaScript libraries that use an API to define property values on an object based on a given path. The function that is generally affected contains this signature: theFunction(object, path, value)

If the attacker can control the value of “path”, they can set this value to `_proto_.myValue. myValue` is then assigned to the prototype of the class of the object.

### Minimist
minimist is a parse argument options module.
Affected versions of this package are vulnerable to Prototype Pollution. The library could be tricked into adding or modifying properties of `Object.prototype` using a `constructor` or `__proto__` payload.
### Poc
```
require('minimist')('--__proto__.injected0 value0'.split(' '));
console.log(({}).injected0 === 'value0'); // true

require('minimist')('--constructor.prototype.injected1 value1'.split(' '));
console.log(({}).injected1 === 'value1'); // true
```
### Types of Attacks
Denial of Service (**DoS**) describes a family of attacks, all aimed at making a system inaccessible to its original and legitimate users. There are many types of DoS attacks, ranging from trying to clog the network pipes to the system by generating a large volume of traffic from many machines (a Distributed Denial of Service - DDoS - attack) to sending crafted requests that cause a system to crash or take a disproportional amount of time to process.

The Regular expression Denial of Service (**ReDoS**) is a type of Denial of Service attack. Regular expressions are incredibly powerful, but they aren't very intuitive and can ultimately end up making it easy for attackers to take your site down.
Examples of Regex:
`regex = /A(B|C+)+D/`
This regular expression accomplishes the following:

- **A** The string must start with the letter 'A'
- **(B|C+)+** The string must then follow the letter A with either the letter 'B' or some number of occurrences of the letter 'C' (the + matches one or more times). The + at the end of this section states that we can look for one or more matches of this section.
- **D** Finally, we ensure this section of the string ends with a 'D'
- The expression would match inputs such as **ABBD, ABCCCCD, ABCBCCCD and ACCCCCD**

It most cases, it doesn't take very long for a regex engine to find a match like this:
`$ time node -e '/A(B|C+)+D/.test("ACCCCCCCCCCCCCCCCCCCCCCCCCCCCD")'
0.04s user 0.01s system 95% cpu 0.052 total`
Most Regex engines will work very similarly (with minor differences). The engine will match the first possible way to accept the current character and proceed to the next one. If it then fails to match the next one, it will backtrack and see if there was another way to digest the previous character. If it goes too far down the rabbit hole only to find out the string doesn’t match in the end, and if many characters have multiple valid regex paths, the number of backtracking steps can become very large, resulting in what is known as catastrophic backtracking.

### Node Addon Api
node-addon-api is a This module contains header-only C++ wrapper classes which simplify the use of the C based N-API provided by Node.js when using C++. It provides a C++ object model and exception handling semantics with low overhead.
Affected versions of this package are vulnerable to **Remote Memory Exposure** via the `napi_get_value_string_utf8`, `napi_get_value_string_latin1 `and` napi_get_value_string_utf16 functions`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2537,9 +2537,9 @@
       "dev": true
     },
     "@types/minimist": {
-      "version": "1.2.0",
-      "resolved": "https://registry.npmjs.org/@types/minimist/-/minimist-1.2.0.tgz",
-      "integrity": "sha1-aaI6OtKcrwCX8G7aWbNh7i8GOfY=",
+      "version": "1.2.3",
+      "resolved": "https://registry.npmjs.org/@types/minimist/-/minimist-1.2.3.tgz",
+      "integrity": "sha512-+bMdgqjMN/Z77a6NlY/I3U5LlRDbnmaAk6lDveAPKwSpcPM4tKAuYsvYF8xjhOPXhOYGe/73vVLVez5PW+jqhw==",
       "dev": true
     },
     "@types/mocha": {
@@ -3095,9 +3095,9 @@
       }
     },
     "ajv": {
-      "version": "6.10.2",
-      "resolved": "https://registry.npmjs.org/ajv/-/ajv-6.10.2.tgz",
-      "integrity": "sha512-TXtUUEYHuaTEbLZWIKUr5pmBuhDLy+8KYtPYdcV8qC+pOZL+NKqYwvWSRrVXHn+ZmRRAu8vJTAznH7Oag6RVRw==",
+      "version": "6.12.3",
+      "resolved": "https://registry.npmjs.org/ajv/-/ajv-6.12.3.tgz",
+      "integrity": "sha512-4K0cK3L1hsqk9xIb2z9vs/XU+PGJZ9PNpJRDS9YLzmNdX6jmVPfamLvTJr0aDAusnHyCHO6MjzlkAsgtqp9teA==",
       "requires": {
         "fast-deep-equal": "^2.0.1",
         "fast-json-stable-stringify": "^2.0.0",
@@ -9671,7 +9671,7 @@
         "external-editor": "^3.0.3",
         "figures": "^3.0.0",
         "lodash": "^4.17.19",
-        "mute-stream": "0.0.8",
+
         "run-async": "^2.4.0",
         "rxjs": "^6.6.0",
         "string-width": "^4.1.0",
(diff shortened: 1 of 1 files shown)


FrancescoBorzi:
thanks @MKokeshi !

FrancescoBorzi (approved):

FrancescoBorzi:
@MKokeshi something weird is happening.. the build is now failing on master after this change, and for some strange reason it didn't run in the PR (I thought it did, but it didn't)

Issue #966, opened 2020-11-14
[BUG] spawnMask


Winfidonarleyan:
If select all `spawnMask` Total value: `-1`
![image](https://user-images.githubusercontent.com/18329778/99133458-aefb3200-264c-11eb-8420-aced7483bbf4.png)

Value `-1` don't support in table `creature`

```sql
mariadb> DELETE FROM `creature` WHERE (`id` = 16980) AND (`guid` IN (1971311));
INSERT INTO `creature` (`guid`, `id`, `map`, `zoneId`, `areaId`, `spawnMask`, `phaseMask`, `modelid`, `equipment_id`, `position_x`, `position_y`, `position_z`, `orientation`, `spawntimesecs`, `wander_distance`, `currentwaypoint`, `curhealth`, `curmana`, `MovementType`, `npcflag`, `unit_flags`, `dynamicflags`, `ScriptName`, `VerifiedBuild`) VALUES
(1971311, 16980, 533, 0, 0, -1, 1, 0, 0, 3749.68, -5114.06, 142.115, 2.93215, 604800, 0, 0, 1, 0, 0, 0, 0, 0, '', 0);
Query OK, 1 row affected (0.00 sec)

1264 - Out of range value for column 'spawnMask' at row 1
```

PR #971, opened 2020-11-14
fix(flags-selector): do not use -1 to select all flags

FrancescoBorzi:
- Closes https://github.com/azerothcore/Keira3/issues/966


Kitzunu:
Looks nice, then we can add a quick option to mark all bits in a bitmax

Helias (approved):

PR #1029, opened 2021-01-04
fix(InhabitType): Update AIR -> FLYING to match wiki

Kitzunu:


src/app/shared/constants/flags/inhabit-type.ts
```diff
@@ -3,6 +3,6 @@ import { Flag } from '../../types/general';
 export const INHABIT_TYPE: Flag[] = [
   { bit: 0, name: 'GROUND' },
   { bit: 1, name: 'WATER' },
-  { bit: 2, name: 'AIR' },
+  { bit: 2, name: 'FLYING' },
```
FrancescoBorzi, line 6: are you sure it's not the wiki that should be updated? `AIR` is consistent with `GROUND` and `WATER`.... while `FLYING` is like `WALKING` and `SWIMMING` ...

Kitzunu: Yea.. core also has air

Issue #1074, opened 2021-02-12
Updating anything in Creature Template adds on extra line


Voxstrasza:
Updating anything in Creature Template adds on an extra line in the 'Diff-query' and trying to run it results in error.

Example of Keira3 query when updating faction: 
```
UPDATE `creature_template` SET `faction` = 35, `spell_school_immune_mask` = 'undefined' WHERE (`entry` = 582);
```
Example of what it should be when updating faction:
```
UPDATE `creature_template` SET `faction` = 35 WHERE (`entry` = 582);
```

55Honey:
https://github.com/azerothcore/azerothcore-wotlk/pull/4358/files

Is it related to the new column added here?

Voxstrasza:
Possibly.  Keira hasn't updated with the new line in creature_table it seems.

@Kitzunu why did you break it? 😠

Helias:
Hehe, kitzunu already created a PR lot time ago for this,but we need to finish it 😅

Unfortunately, I am very busy in this period and I didn't finish it

FrancescoBorzi:
@Kitzunu can't we just add that new field in `creature_template` to make Keira3 work with latest AC master ? I guess we don't need to do everything at once

Kitzunu:
I dont have time atm. But anyone can just copy that part from the open pr

Kitzunu:
feel free to test the PR

PR #1076, opened 2021-02-15
feat(creature_template): spell_school_immune_mask

Kitzunu:
closes https://github.com/azerothcore/Keira3/issues/1074


Kitzunu:
image
image
image
image

FrancescoBorzi:
@Kitzunu looks good overall, but the tests are failing

Kitzunu:
I dont understand the errors, there are like a million errors unrelated to the PR

FrancescoBorzi:
@Kitzunu it's just a test case that is failing because the expected SQL query does not match with the actual SQL query. Which is normal as you added an extra field. So it's just a matter of putting the correct expected query. I can help with it later if needed.

Kitzunu:
nice

Helias (approved):

PR #1122, opened 2021-03-12
Authentication protocol

claudiodfc:
### Added instructions for the ones having problems with this issue using Keira3:

Client does not support authentication protocol requested by server; consider upgrading MySQL client

More information about it here:

https://stackoverflow.com/questions/50093144/mysql-8-0-client-does-not-support-authentication-protocol-requested-by-server


FrancescoBorzi (approved):

FrancescoBorzi:
thanks @claudiodfc

PR #1145, opened 2021-03-25
feat(all-editors): Reload button

FrancescoBorzi:
Suggested by @Stifler82 - New "Reload" button in all editors:

image

image

image

image

### DIFF
--- a/src/app/shared/abstract/service/editors/editor.service.ts
+++ b/src/app/shared/abstract/service/editors/editor.service.ts
@@ -102,7 +102,7 @@ export abstract class EditorService<T extends TableRow> extends SubscriptionHand
     this.reloadEntity(id);
   }
 
-  protected reloadAfterSave() {
+  reloadSameEntity() {
     this.reload(this.loadedEntityId);
   }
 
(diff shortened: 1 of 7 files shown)


src/app/shared/abstract/service/editors/editor.service.ts
```diff
@@ -102,7 +102,7 @@ export abstract class EditorService<T extends TableRow> extends SubscriptionHand
     this.reloadEntity(id);
   }
 
-  protected reloadAfterSave() {
+  reloadSameEntity() {
```
Helias, line 105: void return type

src/app/shared/abstract/service/editors/single-row-complex-key-editor.service.ts
```diff
@@ -83,7 +83,7 @@ export abstract class SingleRowComplexKeyEditorService<T extends TableRow> exten
     this.reloadEntity();
   }
 
-  protected reloadAfterSave() {
+  reloadSameEntity() {
```
Helias, line 86: void return type

src/app/shared/modules/query-output/query-output.component.ts
```diff
@@ -41,4 +49,24 @@ export class QueryOutputComponent<T extends TableRow> {
       this.executeQuery.emit(this.editorService.diffQuery);
     }
   }
+
+  reload() {
```
Helias, line 53: void return type

Helias (changes requested):

Helias (approved):

PR #1152, opened 2021-03-29
Add AUR link

wizzymore:
Add AUR download link for users awareness, I made this a long time ago but I didn't published it anywhere, I think it would be useful for some folks as Manjaro is a very popular linux distro.


FrancescoBorzi:
@wizzymore thanks, is it always updated with the latest Keira3 version?

Helias:
thanks a lot, the arch linux community will be happy after this :grin:

Helias (approved):

PR #1189, opened 2021-04-28
chore(spell-dbc): update base

FrancescoBorzi:


src/app/shared/testing/page-object.ts
```diff
@@ -71,6 +71,17 @@ export abstract class PageObject<ComponentType> {
     this.fixture.detectChanges();
   }
 
+  getLabel(fieldName: string) {
```
Helias, line 74: return type

src/app/shared/testing/page-object.ts
```diff
@@ -71,6 +71,17 @@ export abstract class PageObject<ComponentType> {
...
+    return this.query<HTMLLabelElement>(`.control-label[for="${fieldName}"]`);
+  }
+  getInput(fieldName: string) {
```
Helias, line 77: same here

PR #1209, opened 2021-05-07
Add most tooltips to spells Base tab.

Azcobu:
This adds most (but not all) tooltips to Keira3's spells Base tab.

DB fields that I couldn't find definitive information for are left as TODO, on the basis that no information is better than wrong information.


Helias (approved):

FrancescoBorzi:
I wonder why the CI doesn't run 🤔

FrancescoBorzi (approved):

PR #1229, opened 2021-05-16
feat(GO): gameobject spawn addon editor

FrancescoBorzi:


src/app/config/app-routing.module.ts
```diff
@@ -80,6 +80,7 @@ import { MailLootTemplateComponent } from '../features/other-loots/mail-loot/mai
 import { MailLootHandlerService } from '../features/other-loots/mail-loot/mail-loot-handler.service';
 import { SelectSpellComponent } from '../features/spell/select-spell/select-spell.component';
 import { SpellDbcComponent } from '../features/spell/spell-dbc/spell-dbc.component';
+import { GameobjectSpawnAddonComponent } from '../features/gameobject/gameobject-spawn-addon/gameobject-spawn-addon.component';
```
Kitzunu, line 83: Better use GameobjectAddon instead of "spawn" to match DB name imo

FrancescoBorzi: I only made this exception with `creature`, `gameobject` and their `_addon` table in order to make extra clear that it's about spawns and not the creature/gameobject itself. So this follows the same pattern for consistency :)

Kitzunu: True, but for not "spawn" you would have GameobjectTemplateAddon, no?

FrancescoBorzi: yes

Kitzunu: is whatever I guess

Helias (approved):

PR #1279, opened 2021-06-02
fix(gossip-menu): starting ID when creating custom gossip

FrancescoBorzi:


Stifler82 (changes requested):
`export const GOSSIP_MENU_CUSTOM_STARTING_ID = 65_535;`

FrancescoBorzi:
> 65_535

no, that's the MAX integer that the table can take, while `GOSSIP_MENU_CUSTOM_STARTING_ID` is the number to START with custom IDs. So custom gossips will be 62000, 62001, 62002, etc...

Stifler82:
ah, ok..

PR #1281, opened 2021-06-03
feat: update to ngx-highlightjs 4 and highlightjs 10 + misc improvements

FrancescoBorzi:
This moves our app one step forward.

- replaced deprecated highlights 9.x with latest 10.x
- update ngx-highlighjs to latest 4.x
- remove bad `.forRoot()` used in child modules
- from now on, we will not use highlighjs in unit tests


src/app/config/highlight.config.ts
```diff
@@ -1,9 +1,8 @@
-import sql from 'highlight.js/lib/languages/sql';
+import { HighlightOptions } from 'ngx-highlightjs/lib/highlight.model';
```
MurhafSousli, line 1: You don't need deep import here, just import it from `ngx-highlightjs`

src/app/shared/modules/selectors/area-selector/area-selector.module.ts
```diff
@@ -4,22 +4,14 @@ import { ReactiveFormsModule } from '@angular/forms';
...
   declarations: [AreaSelectorBtnComponent, AreaSelectorModalComponent],
-  imports: [
-    BrowserModule,
-    ReactiveFormsModule,
-    NgxDatatableModule,
-    ModalModule.forRoot(),
-    HighlightModule.forRoot(highlightOptions),
-    SearchButtonsModule,
-  ],
+  imports: [BrowserModule, ReactiveFormsModule, NgxDatatableModule, ModalModule.forRoot(), HighlightModule, SearchButtonsModule],
```
MurhafSousli, line 14: `ModalModule.forRoot()` you should not call forRoot in a reusable module, this should only be called once, if you are importing area-selector.module in multiple places, you should remove `forRoot`

FrancescoBorzi:
@MurhafSousli thanks a lot for your help. I managed to get it compiled, but now it is failing at runtime with:

image

any hint?

MurhafSousli:
@FrancescoBorzi I don't how you use it, I suggest you install the latest hljs v10 (not 11)... load the library only once in the root module and in all other lazy / child modules just import `HighlightModule`

Helias (approved):

Issue #1373, opened 2021-07-24
Add detection_range field in creature_template editor


Helias:
https://github.com/azerothcore/wiki/pull/609

Helias:
Same for ExperienceModifier https://github.com/azerothcore/azerothcore-wotlk/pull/7095

Krejza9:
To be added after merging into master

Krejza9:
@Helias done , you can merge this hotfix for detection_range

PR #1375, opened 2021-07-24
add detection_range to modifiers

Krejza9:
Add detection_range to modifiers
Close: #1373


FrancescoBorzi (approved):

FrancescoBorzi:
@Krejza9 thanks for this PR. Unfortuantely there is still a broken test, can you please fix it? If you need any assistance, just pke me

Krejza9:
I think I found where the problem is, the tests don't count the new column in the DB, I rewrite them

Krejza9:
To be honest, I don't understand why the test is failing.
image

I've got it right there
image
and he skips it like the value isn't even there.

Helias:
I'll have a look

Helias:
the number "20" is auto-generated from `changeAllFields()`, but if you set it as the default value of detection_range it will not trigger the "change" in the unit-test because it is already 20 as value.
I set "1" as the default value because in the DB 1 is the default value and changed the unit-test.

Now it should be ok

Krejza9:
I see, thanks for the explanation and correction.

PR #1410, opened 2021-08-19
feat(Creature/npc_vendor): item extended cost parser 

Helias:
Now the extended cost is parsed.

image

image

Creature entries for testing:
- 35577 multiple items
- 33939 honor points + arena points


src/app/features/creature/npc-vendor/npc-vendor.service.ts
```diff
@@ -1,19 +1,72 @@
...
     protected handlerService: CreatureHandlerService,
     public readonly queryService: MysqlQueryService,
+    public readonly sqliteQueryService: SqliteQueryService,
     protected toastrService: ToastrService,
+    private iconService: IconService,
   ) {
     super(NpcVendor, NPC_VENDOR_TABLE, NPC_VENDOR_ID, NPC_VENDOR_ID_2, handlerService, queryService, toastrService);
   }
+
+  private cache: { [key: string]: Promise<string>[] } = {};
```
FrancescoBorzi, line 24: I think one array of `Promise<string>` is enough, we don't need the object with key

FrancescoBorzi (approved):

PR #1465, opened 2021-09-24
Added faction name id to the faction search

Asthalor:
I added a new field on the faction table, so when people want to edit the faction they will find the correct faction id, instead of the database id (which are not the same :P).

This PR however is not done, because i still dont know how to link the acore faction:faction to the new field i created(faction_name_id).

Test of the new change:

image

I changed the SQlite database by doing the alter:

ALTER TABLE factions ADD COLUMN faction_name_id INTEGER NOT NULL DEFAULT '0';

I dont know if the database changes have a proper pull request, but the gitHub commit added the whole database. So if you want to know what changed it was that.

Please let me know how to link it or anything you see that can be improved,

### DIFF
--- a/src/app/shared/types/faction.type.ts
+++ b/src/app/shared/types/faction.type.ts
@@ -2,9 +2,10 @@ import { TableRow } from './general';
 
 export const FACTION_TABLE = 'factions'; // sqlite
 export const FACTION_ID = 'm_ID';
-export const FACTION_SEARCH_FIELDS = [FACTION_ID, 'm_name_lang_1'];
+export const FACTION_SEARCH_FIELDS = [FACTION_ID, 'faction_name_id', 'm_name_lang_1'];
 
 export class Faction extends TableRow {
   m_ID: number = 0;
   m_name_lang_1: string = '';
+  faction_name_id: number = 0;
 }
--- a/src/app/shared/modules/selectors/faction-selector/faction-selector-modal.component.html
+++ b/src/app/shared/modules/selectors/faction-selector/faction-selector-modal.component.html
@@ -7,6 +7,9 @@
         <div class="form-group col-3">
           <input [formControlName]="'m_ID'" type="number" class="form-control form-control-sm" id="m_ID" placeholder="ID" />
         </div>
+        <div class="form-group col-3">
+          <input [formControlName]="'faction_name_id'" class="form-control form-control-sm" id="faction_name_id" placeholder="Faction" />
+        </div>
         <div class="form-group col-5">
           <input [formControlName]="'m_name_lang_1'" class="form-control form-control-sm" id="m_name_lang_1" placeholder="Faction Name" />
         </div>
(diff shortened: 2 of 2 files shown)


Helias:
Checked, you need to add the value of the faction, because now they are all 0 (you can take the value from faction dbc data) and, selecting the value in the faction selector the input should be filled with the "faction value" and not with the "faction ID"

src/app/shared/types/faction.type.ts
```diff
@@ -2,9 +2,10 @@ import { TableRow } from './general';
 
 export const FACTION_TABLE = 'factions'; // sqlite
 export const FACTION_ID = 'm_ID';
-export const FACTION_SEARCH_FIELDS = [FACTION_ID, 'm_name_lang_1'];
+export const FACTION_SEARCH_FIELDS = [FACTION_ID, 'faction_name_id', 'm_name_lang_1'];
```
Helias, line 5: here you need to change the value of the faction_ID to select the faction_name_id instead of the m_ID when the user select a faction through the faction selector

```typescript
export const FACTION_ID = 'faction_name_id';
export const FACTION_SEARCH_FIELDS = ['m_ID', 'faction_name_id', 'm_name_lang_1'];
```

Asthalor: The problem is that on the faction text field we need to have the m_ID, because its the same as the info we get from the sniffs.
Im going to add the data using the sqlitebrowser and see how it goes. If i have to change something, with real data will be easier i think.

Asthalor:
I managed to add the data to the factions table. I will add the SQL as a commit.
I used the query 
`SELECT template.id, template.Faction, faction.Name_Lang_enUS  FROM factiontemplate_dbc template, faction_dbc faction where template.Faction = faction.ID;` to search for the data to populate.
I used a delete + insert. Is costly, but its going to be executed once. Anyway i think that with those changes the view is way better than before.
Example of Tahurrisan now:

image

Helias:
`Factions update.sql` is not needed.
For the rest, I tested it, it seems working

Asthalor:
Awesome. i thought that the values were wrong, but if it is tested and seems good its all good.

Helias:
Not sure but I think that you pushed the sqlite version, let me check again

Asthalor:
Yes, on the files changed i added the sqlite version with the column faction name id.

Asthalor:
The `Factions update.sql` is used to populate the new field added to the sqLite database if it helps.

Helias:
you can remove the faction update sql if you pushed the new sqlite version, it's redundant

Asthalor:
> I would suggest to populate the data and push the sqlite with them, so the end-user does not need to import it

Thats a good idea. Let me add a commit for that

Asthalor:
Ok! Thanks for the advise. It should be all good now!

Helias:
Ok, let's wait the pipeline to succeed and I will merge it!

PR #1572, opened 2021-12-09
Modify information in README.md about MySQL

pangolp:
- The SQL statement is modified.
- Field information is added.

Currently, when using MySQL 8.x, it is necessary to change the encryption as specified in the readme, however, the connection type must be clarified, which is what I am adding. Take this opportunity to upload a detail of what the values mean within the SQL statement.


Helias (approved):

FrancescoBorzi:
thanks @pangolp for your contribution!

Issue #1695, opened 2022-02-12
Sidebar icons have something wrong


FrancescoBorzi:
on macOS you can see some squares instead of the icons

buddiman:
Possibly related?
![image](https://user-images.githubusercontent.com/10009755/153761563-972e41ab-7bb9-4b86-bbc1-ed78c9ed0678.png)

The squares where the arrows should be even have the correct animation:
![Arrow](https://user-images.githubusercontent.com/10009755/153761649-0a49c69d-b801-4bf8-8b1c-3bb99c5c2d67.gif)

Is it the same visual bug? I'm using the latest Windows release.

FrancescoBorzi:
@buddiman yes

PR #1779, opened 2022-04-29
fix: adapt sidebar icon styling to fontawesome 6

buddiman:
Closes https://github.com/azerothcore/Keira3/issues/1695  

It seems that Font Awesome was upgraded from v5 to v6. Version 6 introduced breaking changes on how to use the icons with pseudo-elements. You can find the changes under [this link](https://fontawesome.com/docs/web/setup/upgrade/pseudo-elements), especially the paragraph "Updating Font Rule Syntax". Normally there should be an automated backward compatibility feature, but it seems this didn't work here.

The result after the changes is the following:
image


Helias (approved):

Helias:
Thanks for this PR!

Issue #1086, opened 2021-02-23
Save the sql code temporary in Keira3 via "To Be Commited" button


Helias:

FrancescoBorzi:
suggestion by @Stifler82

Stifler82:
In practice, create a "To Be Committed" button that creates within another section of Keira all the changes that are made (in that Keira session), and that solves the SQL code as briefly as possible and that summarizes all queries of all miscellaneous changes that have been made so, GameObject, NPC, Item, etc, (if) have been changed.

Issue #741, opened 2020-06-05
Ability to export whole entity


FrancescoBorzi:
E.g. export whole Creature's SQL

Issue #401, opened 2020-02-27
[Feature Request] Save as .sql


Kitzunu:
It might be cool to have but don't know exactly how it would be done.

Perhaps either:
1. SAI for example: Have an option to save the SAI as "Anachronos - SAI - DATE.sql" or something
2. Have an option to store all changes you make across everything into a file that you then can name whatever you want. For example if you do an update of several things that link together. For example Updating some creature template stuff, some SAI stuff, some Condition stuff etc.

Yea I am a lazy fucker xD Copy pasta takes time :PPP

Amazing work so far btw <3

Kitzunu:
Or maybe, being able to save to an existing file.

For example, being able to open an SQL file in Keira (Fex opening a generated SQL update file and being able to import all fixes directly to it)

sudlud:
#1086 related to 2.

PR #1781, opened 2022-05-12
Feat/export sql script

buddiman:
## Proposed features
This is a first draft how to implement https://github.com/azerothcore/Keira3/issues/401  

I created a rudimentary export functionality, I created a PR already as I'm open to more input/suggestions/requirements.
At the moment, this only works for the *creature_template* editor.

To test it, enter a path to a (non)existing SQL file in the new textbox in the dashboard:  
image
On Unix systems you have to manually edit the path in the textbox for now as I used the `\` character

I created a new service for handling the functionality as it may can be used later for other features.

## Ideas
- Implement https://github.com/azerothcore/Keira3/issues/741
- Directly run the bash script from AC to create a new SQL update file 
- Maybe this can be used for https://github.com/azerothcore/Keira3/issues/1086

### DIFF
--- a/src/app/shared/modules/query-output/query-output.component.ts
+++ b/src/app/shared/modules/query-output/query-output.component.ts
@@ -16,6 +16,7 @@ export class QueryOutputComponent<T extends TableRow> extends SubscriptionHandle
   @Input() docUrl: string;
   @Input() editorService: EditorService<T>;
   @Output() executeQuery = new EventEmitter<string>();
+  @Output() saveToSQLFile = new EventEmitter<string>();
   selectedQuery: 'diff' | 'full' = 'diff';
   private modalRef: BsModalRef;
 
@@ -43,6 +44,14 @@ export class QueryOutputComponent<T extends TableRow> extends SubscriptionHandle
       this.executeQuery.emit(this.editorService.fullQuery);
     } else {
       this.executeQuery.emit(this.editorService.diffQuery);
+    }
+  }
+
+  saveToSQL(): void {
+    if (this.showFullQuery()) {
+      this.saveToSQLFile.emit(this.editorService.fullQuery);
+    } else {
+      this.saveToSQLFile.emit(this.editorService.diffQuery);
     }
   }
 
--- a/src/app/shared/services/save-query.service.ts
+++ b/src/app/shared/services/save-query.service.ts
@@ -1,0 +1,19 @@
+import { Injectable } from '@angular/core';
+import * as fs from 'fs';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class SaveQueryService {
+  filePath: string = '';
+
+  appendToFile(query: string) {
+    fs.appendFile(this.filePath, query + '\n', { flag: 'a+' }, (err) => {
+      if (err) console.log(err);
+    });
+  }
+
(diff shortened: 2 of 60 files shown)


src/app/shared/modules/query-output/query-output.component.ts
```diff
@@ -46,6 +47,14 @@ export class QueryOutputComponent<T extends TableRow> extends SubscriptionHandle
     }
   }
 
+  saveToSQL(): void {
+    if (this.showFullQuery()) {
+      this.saveToSQLFile.emit(this.editorService.fullQuery);
+    } else {
+      this.saveToSQLFile.emit(this.editorService.diffQuery);
+    }
```
Helias, line 55:
```suggestion
    const fullOrDiffQuery = this.showFullQuery() ? this.editorService.fullQuery : this.editorService.diffQuery);
    this.saveToSQLFile.emit(fullOrDiffQuery);
```

src/app/shared/services/save-query.service.ts
```diff
@@ -0,0 +1,19 @@
...
+export class SaveQueryService {
+  filePath: string = '';
+
+  appendToFile(query: string) {
+    fs.appendFile(this.filePath, query + '\n', { flag: 'a+' }, (err) => {
+      if (err) console.log(err);
+    });
+  }
+
+  setFilePath(path: string) {
```
Helias, line 16:
```suggestion
  setFilePath(path: string): void {
```

src/app/shared/services/save-query.service.ts
```diff
@@ -0,0 +1,19 @@
+import { Injectable } from '@angular/core';
+import * as fs from 'fs';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class SaveQueryService {
+  filePath: string = '';
```
Helias, line 8: if the type is inferred, we usually avoid writing it explicitly
```suggestion
  filePath = '';
```

src/app/shared/services/save-query.service.ts
```diff
@@ -0,0 +1,19 @@
...
+
+  appendToFile(query: string) {
+    fs.appendFile(this.filePath, query + '\n', { flag: 'a+' }, (err) => {
+      if (err) console.log(err);
```
Helias, line 12:
```suggestion
      if (err) {
        console.log(err);
      }
```

Helias:
Thanks for your PR, I don't know if you will take care of the unit-test of your service but they are really important to do for the Keira3 project.

Moreover, is the path can be chosen when the user click on "save to sql"? and is there a way to easily choose a path and make it as default without writing it?

buddiman:
> Thanks for your PR, I don't know if you will take care of the unit test of your service but they are really important to do for the Keira3 project.

It's basically a WIP at the moment, I only already opened the PR to get early feedback on what is needed/wished.
I'll do unit tests for sure.

> Moreover, is the path can be chosen when the user clicks on "save to SQL"? and is there a way to easily choose a path and make it as default without writing it?
I want to change the textbox in the dashboard so that you can open a "file explorer dialog" as electron already provides this feature. As for having a default path, that would be a good idea.

FrancescoBorzi:
closing as it looks like abandoned

PR #1814, opened 2022-06-19
Starting Translation to French

Plep-m:


Helias:
Merged, even if incomplete, but it will be available in the next Keira3 release.

PR #1817, opened 2022-06-19
i18n: Spanish translation for SmartAI and SpellDBC sections

ZeegCL:
As requested by @Helias , I'm starting from the bottom to avoid collitions with other collaborators. This is like 17% of the file, will continue later with the rest of the sections.


Helias:
will you complete it or I can merge this?

ZeegCL:
> will you complete it or I can merge this?

I'll do it by sections, so @pangolp can merge his updates too. You are safe to merge this PR

pangolp:
> I'll do it by sections, so @pangolp can merge his updates too. You are safe to merge this PR

If you take care of the translations, the truth is that it's better for me, because this way I can take the opportunity to really look at other things. But anyway, I think it's good that you do it in parts, it's not necessary to merge all the texts together at the beginning, because in the long run, some of them will surely change. Besides, there are words that should remain in English, such as: gossip, quest, npc, creature, e.t.c, because that way you can relate them to their respective table in the database or in the script.

pangolp:
> will you complete it or I can merge this?

For me, the best thing to do is to merge it, because the file is very large, so we can see if more people are encouraged to translate it too, and if not, over time, we will make translations, as needed, but if it is for me, I apply the merge.

PR #1821, opened 2022-06-20
Update ru.json

Meramaf:


Helias:
Merged, even if incomplete, but it will be available in the next Keira3 release.

PR #1827, opened 2022-06-23
Update pt.json

Gultask:
A bit rough around the edges. Started with this because no one spoke up in the Discord :)
May be inaccurate at several points because of my inexperience with Keira and WoW emulation in general.
Had some trouble in Spell Effects section (don't know what the word 'according' means in those cases) and another couple of lines that I couldn't understand at all.

Biased towards brazilian portuguese and uses pre-localized-era MMO slang language such as turning english words into conjugated portuguese verbs (e.g. Cast -> eu casto, você casta, nós castamos) that I remember from childhood memories. It was either that or keep it in english :P 

Better than nothing, I suppose. I'm not particularly good with portuguese and I didn't necessarily get top marks in portuguese class. Quite the opposite, in fact. I hated grammar. So I'm open to reviews, and if anyone wants, they're free to take over. Otherwise it's completed as far as I can make it.


Helias:
is it finished?

Gultask:
Did one last review because I missed a couple of lines.

Should be good to go now. It's finished. :)

Helias:
Thank you!

PR #2054, opened 2022-12-25
feat(deps): Angular 15, ngx-bootstrap 10

FrancescoBorzi:


.github/workflows/build-linux.yml
```diff
@@ -31,6 +31,6 @@ jobs:
             ${{ runner.OS }}-node-
             ${{ runner.OS }}-
       - name: Install dependencies
-        run: npm ci
+        run: npm ci --legacy-peer-deps
```
FrancescoBorzi, line 34: this can be reverted once we get rid of ngx-typsafe-forms

PR #2083, opened 2023-01-07
feat: add 3d-model-viewer for NPCs, GameObjects and improve the item one

Helias:
image

image

### DIFF
--- a/src/app/features/creature/creature-template/creature-template.component.html
+++ b/src/app/features/creature/creature-template/creature-template.component.html
@@ -1,6 +1,6 @@
 <keira-top-bar [handler]="handlerService"></keira-top-bar>
 
-<div class="container-fluid">
+<div class="container-fluid editor-with-preview" [ngClass]="{ 'show-preview': showItemPreview }">
   <span *ngIf="editorService.loading" [translate]="'LOADING'"></span>
 
   <div *ngIf="editorService.form && !!editorService.loadedEntityId && !editorService.loading">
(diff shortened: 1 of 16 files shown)


angular.json
```diff
@@ -127,6 +127,7 @@
             "karmaConfig": "src/karma.conf.js",
             "codeCoverage": true,
             "codeCoverageExclude": [
+              "src/app/features/model-3d-viewer/helper.ts",
```
FrancescoBorzi, line 130: I'd prefer `/* istanbul ignore file */` inside the file instead of this

src/app/features/creature/creature-template/creature-template.component.html
```diff
@@ -535,5 +535,20 @@
...
+
+    <div class="preview-container" [ngClass]="showItemPreview ? 'show-preview' : 'hide-preview'">
+      <perfect-scrollbar class="transparent">
+        <div class="creature-preview">
+          <keira-model-3d-viewer
+            [viewerType]="NPC_VIEWER_TYPE"
+            [modelId]="editorService.form.controls.modelid1.value"
+          ></keira-model-3d-viewer>
+        </div>
+      </perfect-scrollbar>
```
FrancescoBorzi, line 547: if it's possible, I would avoid  using `<perfect-scrollbar>`  because it's deprecated and we should remove it completely from Keira3

src/app/features/gameobject/gameobject-template/gameobject-template.component.html
```diff
@@ -91,5 +91,20 @@
         </div>
       </div>
     </form>
+
+    <div class="preview-container" [ngClass]="showItemPreview ? 'show-preview' : 'hide-preview'">
+      <perfect-scrollbar class="transparent">
```
FrancescoBorzi, line 96: same here

angular.json
Helias: done

FrancescoBorzi (approved):

PR #2084, opened 2023-01-07
fix: SAI generate comment, closes #1820

### LINKED ISSUE #1820: smartAI: generate comments on eventype 61 causes link arugment to be 1
If you generate comments on eventype 61 (link), the argument param1 will become 1, which is false and will make it not run and cause errors. It should not add a 1 in param1. (should be 0)
image

Helias:
closes #1820

### DIFF
--- a/src/app/shared/modules/sai-editor/sai-editor.service.ts
+++ b/src/app/shared/modules/sai-editor/sai-editor.service.ts
@@ -92,7 +92,7 @@ export class SaiEditorService extends MultiRowComplexKeyEditorService<SmartScrip
   }
 
   async generateComments() {
-    for (const row of this._newRows) {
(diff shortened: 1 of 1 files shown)


src/app/shared/modules/sai-editor/sai-editor.service.ts
```diff
@@ -92,7 +92,7 @@ export class SaiEditorService extends MultiRowComplexKeyEditorService<SmartScrip
   }
 
   async generateComments() {
-    for (const row of this._newRows) {
+    for (const row of structuredClone(this._newRows)) {
```
FrancescoBorzi, line 95: it would be nice to have a unit test that would fail before this bug was fixed, if possible please add

Issue #2073, opened 2022-12-29
minor: first switch language is not coherent with the second selector


Helias:
how to reproduce:
- in the connect screen select the languages via select options
- clicking on connect in the dashboard screen will not be shown the language previously selected in the select option dropdown

PR #2086, opened 2023-01-07
fix: switch language persistent

Helias:
closes https://github.com/azerothcore/Keira3/issues/2073

### DIFF
--- a/src/app/shared/modules/switch-language/switch-language.service.spec.ts
+++ b/src/app/shared/modules/switch-language/switch-language.service.spec.ts
@@ -1,0 +1,42 @@
+import { NgModule } from '@angular/core';
+import { TestBed, waitForAsync } from '@angular/core/testing';
+import { ModalConfirmComponent } from '@keira-shared/modules/modal-confirm/modal-confirm.component';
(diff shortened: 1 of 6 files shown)


FrancescoBorzi (approved):

src/app/shared/modules/switch-language/switch-language.service.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+@NgModule({
+  declarations: [ModalConfirmComponent],
+  imports: [],
+})
+class TestModule {}
+
+describe('SwitchLanguageService', () => {
+  beforeEach(waitForAsync(() => {
+    TestBed.configureTestingModule({
+      imports: [ModalModule.forRoot(), TestModule, TranslateTestingModule],
```
FrancescoBorzi, line 18: why are we creating and importing `TestModule` to declare `ModalConfirmComponent` while we could just add it inside the `declarations: [ .... ]` array here?

Helias: removed, it was not used

FrancescoBorzi (approved):

PR #2145, opened 2023-02-09
Updates and translates the "es" (Spanish) json file

M4th3m4tic4l:
I finish with all the missing translations that were left pending in the 'es' file.

Here is an example (I have edited a couple of bugs in some translations that I detected while testing the application, so some things have changed since I made these screenshots):

misc01
español01

Although I didn't translate **everything** that needed to be translated, when I was testing for a while I realized that there were sections that were not available in the file, for example, the 'flags' of some checkbox options, or the descriptive texts of them, etc.

I guess they are separate things (correct me if I'm wrong)

In any case, I leave here this contribution, which has taken me quite a long time to make (no kidding), and I don't think I will regret having done it. After all, I have done it for the benefit of someone who does not understand English well 🧙


Helias (approved):

Helias:
Thanks a lot!

Issue #1741, opened 2022-03-12
Sulans list of some tooltip changes for Keira3(Last updated 20 January 2023)


Sulan2:
I will update this post if I find something in the future

SmartAI Editor

Event -> 102 Near Player Negation
The I Icon next to Near_Player_Negation should say
"Amount of players needed within a range to STOP the action"

Param 1 (maxPlayers) should be called "Amount of Players" and the I icon should say "Players needed to STOP the Action"

Event -> 101 Near Player
The I Icon next to Near_Players should say
"Amount of Players needed within a range to Start the action"

Param 1 (minPlayers) should be called "Amount of Players" and the I icon should say "Players needed to Start the Action"

Action -> 69 MOVE_TO_POS 
Param 3 (isControlled)
The I icon should say
"0 = aggro/threat will make it focus on player instead"
"1 = Fixated to run to the xyzo position no matter what"

Target -> 5 Hostile Random
Param 1 2 and 3 have a meaning here (The world server says it)
Param 2 is "Player Only" Boolean 0 or 1
Param 1 and 3 I dont know what they do however but a guess is that one of em is about distance

Edit: (All target Hostile seem to use this, Param 2 for target Hostile 3 4 5 6 are all player only boolean)

Target -> 19 Closest Creature
The I Icon for MustBeDead
This is just a tooltip error currently using 1 for both. I dont know what value of 0 or 1 correlated to what here

----------------------------------
Creature Template

Type_flags -> BOSS_MOB

The description here should say
"Hidden level in Creature Portrait (Shown as ??) Immune to Knockback. Always deal full damage to players, doesnt miss spells."

Type -> 9 Mechanical
This doesnt have any description for anything, maybe its not such a big deal but u can add 
"(cannot be healed by healing spells)" here to Mechanical if u want

-------
Item Template

Spell -> Trigger_1_2_3_4_5
These could have an I icon that says
0 = Use, uses Charges and cooldown below
1 = Equip, uses Cooldown Column in spell_proc_event table database
2 = Chance on Hit, uses ppmrate and cooldown below
5 = Having this item in your bag or equipped will also trigger an aura on you.

Sulan2:
On this site https://www.azerothcore.org/wiki/smart_scripts

It has flag values and meaning for Castflags and Summon Type, wouldnt it be possible to get them into Keira3 directly?

In SmartAI Editor 
Action -> 11 Cast -> CastFlags

1 | Interrupt any spell casting
[...]

PR #2156, opened 2023-02-18
fix: tooltip description for event 102 near players negation, closes #1705

### LINKED ISSUE #1705: Some thoughts on SmartAI Editor - Event 102 Near Players Negation
I feel like the info text on this event isnt really clear/ is wrong

I created a NPC in SmartAi editor that, when NO players are in range it will trigger the event function
In order to make this work in game, Event is 102 Near Players Negation and Maxplayers is set to 1

However  the info on maxplayers says the value choosen is = whats needed to trigger the event
Which means if I set the value of maxplayers to 1 the event should trigger if 1 player is close

But shouldnt the info on Maxplayers read "Amount of players needed for the event to NOT trigger"? because thats how it works in game
Setting maxplayers at 1 will make it so that 1 player in range will STOP the event from happening

The way both events 101 and 102 are worded now feels like they are doing the same thing but I think the 102 event arent inversed in a maxplayer/minplayers kind of way
instead 102 is the amount of players needed for an action to NOT happen
101 is the amount of players needed for an action to happen

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://app.bountysource.com/issues/106105575-some-thoughts-on-smartai-editor-event-102-near-players-negation?utm_campaign=plugin&utm_content=tracker%2F139701226&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://app.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F139701226&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Helias:
closes https://github.com/azerothcore/Keira3/issues/1705


Helias:
duplicate of https://github.com/azerothcore/Keira3/issues/1741

PR #2178, opened 2023-02-27
Change misleading description in sai-actions.ts

Gultask:
Saw it in action and tested the event ingame. Setting both to 0 removes the effect and does not give an error.


Helias:
thank you!

PR #2280, opened 2023-04-16
Add new creature extra flag CREATURE_FLAG_EXTRA_DONT_OVERRIDE_SAI_ENTRY

Gultask:
https://github.com/azerothcore/azerothcore-wotlk/pull/15976


avarishd:
Maybe worth adding that IDs should not overlap? I know you've added it in the wiki, but still.

Gultask:
Yeah thought so as well but I don't wanna pollute the flag description too much

avarishd:
> Allows creatures to use both GUID and ENTRY specific SAI (IDs should not overlap)

That is probably fine?

PR #2383, opened 2023-08-05
feat: add icon in spell dbc search

Helias:


src/app/shared/modules/icon/icon.service.ts
```diff
@@ -3,11 +3,16 @@ import { MysqlQueryService } from '@keira-shared/services/mysql-query.service';
 import { SqliteQueryService } from '@keira-shared/services/sqlite-query.service';
 import { map, mergeMap, Observable, of } from 'rxjs';
 
+export const trade_engineering_icon_ID = 1;
```
FrancescoBorzi, line 6: what about uppercase?

FrancescoBorzi (approved):

src/app/shared/modules/icon/icon.service.ts
Helias: done

PR #2586, opened 2023-11-17
feat(deps): Angular 17

FrancescoBorzi:


.github/workflows/windows.yml
```diff
@@ -35,7 +35,7 @@ jobs:
             ${{ runner.OS }}-node-
             ${{ runner.OS }}-
       - name: Install dependencies
-        run: npm ci
+        run: npm ci --legacy-peer-deps
```
Helias, line 38: if you prefer we could make a `.npmrc` file in the root directory with `legacy-peer-deps=true` so we can change this conf from one single file, but in this way is hurtless too

Helias (approved):

Issue #2629, opened 2023-12-25
Searching for something and then deleting the string will create a broken query with %%


FrancescoBorzi:
![image](https://github.com/azerothcore/Keira3/assets/75517/47c33c28-87c4-45a9-ae80-9868f06c0c88)

Issue #1129, opened 2021-03-17
Generated SQL query can be too restrictive


eTzmNcbkrng:
When entering search parameters in the GUI, the generated SQL query can be too restrictive and not return valid entries that have null data in some fields.

I will use the Creature searcher as an example, the first time entering the "Select Creature" window, the fields are empty and the generated SQL query is
```sql
SELECT * FROM `creature_template` LIMIT 50
```

If I put `3103` in the first field and click Search, one of the results is Makrura Clacker, which is the NPC with ID `3103`.
The generated query for this is
```sql
SELECT * FROM `creature_template` WHERE (`entry` LIKE '%3103%') LIMIT 50
```

So far, so good.

If I enter some data in the subname field and then delete it, the query changes to this
```sql
SELECT * FROM `creature_template` WHERE (`entry` LIKE '%3103%') AND (`subname` LIKE '%%') LIMIT 50
```

Makrura Clacker is no longer returned because it has NULL data in the subname.

This bug is complicated by the fact that the search still returns entries that have empty subnames so there must be something unusual about Makrura Clacker.
I've only just started using Keira so haven't had a lot of time to find more entries with this issue, the only other time I've noticed it is with Dark Iron Rifleman (NPC ID `6523`)

It looks as though if data is entered into the 'Entry' field `'entry' LIKE '%*%'` is added to the query, when you delete text from this field it is completely removed from the generated query, if this function were extended to the Name and Subname fields the issue should be resolved.

eTzmNcbkrng:
Makrura Clacker has valid data in the Name field.
![image](https://user-images.githubusercontent.com/26512802/111461262-db04fc00-8781-11eb-9324-4774da755079.png)
Makrura Clacker does not have valid data in the Subname field
![image](https://user-images.githubusercontent.com/26512802/111461313-eb1cdb80-8781-11eb-895b-cac0dc4268c0.png)

PR #2630, opened 2023-12-25
fix(search-query): empty string parameters + small refactoring

FrancescoBorzi:
- Closes https://github.com/azerothcore/Keira3/issues/2629
- Closes https://github.com/azerothcore/Keira3/issues/1129
- Move unit tests of abstract class in a separate spec file
- Move query-services together into the same folder


src/app/shared/services/query/base-query.service.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+      'should correctly work',
+      waitForAsync(async () => {
+        const { service } = setup();
+        const value = 'mock result value';
+        spyOn(service, 'query').and.returnValue(of([{ v: value }]));
+        const query = 'SELECT something AS v FROM my_table WHERE index = 123';
+
+        expect(await service.queryValueToPromise(query)).toEqual(value);
+        expect(service.query).toHaveBeenCalledTimes(1);
+        expect(service.query).toHaveBeenCalledWith(query);
```
Helias, line 38: you could put directly to have been called once with

Helias (approved):

Issue #2758, opened 2024-02-23
Item Editing StatType2 is wrong for mana


StraysFromPath:
It lists value 0 as mana, 1 as health, and 2 does not exist.

In reality value 0 is blank, 1 is health, I think 2 might be mana?

Helias:
I think you're right, the stat type 0 should be "blank/null" and mana should be 2, but I can't find any item in item_template with stat_type* = 2, so it's probably unused.

By the way, I fixed this on https://github.com/azerothcore/Keira3/pull/2761

And I reported it in AoWoW https://github.com/Sarjuuk/aowow/issues/389, hope they can give us a second opinion.

PR #2761, opened 2024-02-24
fix(item-preview): issue with mana stats when statType=0 + statValue

Helias:
closes https://github.com/azerothcore/Keira3/issues/2758


apps/keira/src/app/features/item/item-template/item-preview.service.ts
```diff
@@ -546,9 +546,10 @@ export class ItemPreviewService {
       const type = Number(itemTemplate['stat_type' + i]);
       const qty = Number(itemTemplate['stat_value' + i]);
 
-      if (!qty || !type || type <= 0) {
+      if (!qty || (!type && type !== 0) || type < 0) {
         continue;
       }
+
```
FrancescoBorzi, line 552: missing new unit tests... whenever we fix a bug, ideally we should add a new unit test case that would catch that bug

FrancescoBorzi (approved):

PR #2799, opened 2024-03-07
refactor: migrate to standalone components

FrancescoBorzi:


apps/keira/src/app/features/creature/creature-template/creature-template.component.ts
```diff
@@ -26,12 +26,42 @@ import {
...
 
 @Component({
   changeDetection: ChangeDetectionStrategy.OnPush,
   selector: 'keira-creature-template',
   templateUrl: './creature-template.component.html',
   styleUrls: ['./creature-template.component.scss'],
+  standalone: true,
+  imports: [
+    TopBarComponent,
+    NgClass,
```
Helias, line 50: let's remove ngClass and replace it with [class]

FrancescoBorzi: better in another PR, because this is getting too big already

Issue #2174, opened 2023-02-25
[Bug Report] SAI Editor sometimes fails to generate comments


Gultask:
Open SmartAI tab
Search by entity -> Timed Actionlist -> 2034400
Generate Comments. All comments there are custom so the Generate Comments should update every row. But it does not.

Behaviour is version 3.5.1:

https://user-images.githubusercontent.com/100873791/221333640-b4cbe742-dbc3-46b9-b7aa-d19f8a6bd491.mp4

Helias:
weird, I will investigate
I was thinking to put two buttons "Generate all comments", "Generate comment" so the user can choose to generate the comments for all or only for the selected row

Gultask:
That's a good idea!
It'd also effectively close https://github.com/azerothcore/Keira3/issues/2107 as well

Issue #2107, opened 2023-01-16
[Feature Request] SAI Editor: Ability to "Lock" comments for each row so that it won't generate


Gultask:
When creating SAI scripts I often want to create custom text for the comments to make things clearer. I.e. Instead of "Say Line 0" it's "Say Line 'Hal...'"
However if I want to add another row, and want to generate a comment for it, I want to keep my changes so that I don't need to rewrite it later.

https://user-images.githubusercontent.com/100873791/212588794-38331f2f-26de-4fea-b51e-80cffc1d2fba.mp4

PR #2819, opened 2024-03-17
fix(SAI): trigger diff query when generating comments + refactor

### LINKED ISSUE #2467: SAI Comments not triggering a diff
If you update a SAI script with only Generate comments it will not provide a new diff and the new comment will not be updated. You need to change an actual param for the comment to update in the diff

Helias:
- Closes #2467
- closes https://github.com/azerothcore/Keira3/issues/2174

Hope this fix does not break https://github.com/azerothcore/Keira3/issues/1820, @Gultask if you can test would be much appreciated, I tested it locally and I could not reproduce the issue, but it's better a double check for this


.gitignore
```diff
@@ -42,6 +42,7 @@ apps/keira/coverage
 npm-debug.log
 testem.log
 /typings
+/src/
```
FrancescoBorzi, line 45: ?

FrancescoBorzi, line 45: ?

FrancescoBorzi:
@Helias what about tests?

FrancescoBorzi (approved):

Gultask:
It did fix the problem, ty!

Gultask:
doesn't seem to have a problem with LINK events either

Gultask:
Closes https://github.com/azerothcore/Keira3/issues/2174 as well. Also your suggestion of generating comments for a single row only would be highly appreciated because I was actually starting to like having this bug for that reason 😅

Helias:
do you refer to this https://github.com/azerothcore/Keira3/issues/2107 ? I could work on it

thanks for testing, I am merging this PR and closing the two issues, I will prioritize this issue letting the functionality to generate comment only for the selected row

Issue #1942, opened 2022-10-05
SAI editor: ability to duplicate rows


Sulan2:
Would love if it would be possible to duplicate rows and just change the ID instead of having to go into HediSql to do this

Could be a button next to add new row button or on the right click menu

avarishd:
That would be so nice, could additionally +1 the ID automatically, upon duplication.

PR #2823, opened 2024-03-17
feat(multi-row-editor): duplicate selected row

FrancescoBorzi:
- Closes https://github.com/azerothcore/Keira3/issues/1942

- TODO: test more manually && write integration tests


libs/shared/base-abstract-classes/src/service/editors/multi-row-editor.service.ts
```diff
@@ -187,16 +194,16 @@ export abstract class MultiRowEditorService<T extends TableRow> extends EditorSe
...
-  protected addIdToNewRow(newRow): void {
-    newRow[this._entityIdField] = Number.parseInt(this.loadedEntityId, 10);
+  protected addIdToNewRow(newRow: T): void {
+    newRow[this._entityIdField as keyof T] = Number.parseInt(this.loadedEntityId, 10) as T[keyof T];
   }
 
-  addNewRow(): void {
-    const newRow = new this._entityClass();
+  addNewRow(copySelectedRow = false): void {
+    const newRow: T = copySelectedRow && this.hasSelectedRow() ? { ...this.getSelectedRow() } : new this._entityClass();
```
FrancescoBorzi, line 202: Note for reviewer: this is the actual copy feature

libs/shared/base-abstract-classes/src/service/editors/multi-row-editor.service.ts
```diff
@@ -43,6 +46,25 @@ export abstract class MultiRowEditorService<T extends TableRow> extends EditorSe
     this.initForm();
   }
 
+  private getRowIndex(id: string | number): number {
+    for (let i = 0; i < this._newRows.length; i++) {
+      if (id === this._newRows[i][this._entitySecondIdField]) {
+        return i;
+      }
+    }
```
Helias, line 54: What about using .findIndex?
If the index result is -1 then return 0, otherwise return the index that you found.

libs/shared/base-abstract-classes/src/service/editors/multi-row-editor.service.ts
Helias: not sure if we have to replace `{ ...this.getSelectedRow() }` with a deep clone of the object using `structuredClone(this.getSelectedRow())`

libs/shared/base-abstract-classes/src/service/editors/multi-row-editor.service.ts
FrancescoBorzi: let's refactor it in another PR (this code is not new, I just moved it), there are other little things that can be improved but I don't like to make the PR bigger than the original scope

libs/shared/base-abstract-classes/src/service/editors/multi-row-editor.service.ts
FrancescoBorzi: @Helias why a deep clone? all the properties of each row are always strings or numbers, never objects.

Helias: then it's ok

libs/shared/base-abstract-classes/src/service/editors/multi-row-editor.service.ts
Helias: ok

Helias (approved):

PR #2824, opened 2024-03-17
feat(SAI): generate single comment

### LINKED ISSUE #2467: SAI Comments not triggering a diff

Helias:
- closes https://github.com/azerothcore/Keira3/issues/2107

I've added the possibility to generate a single comment OR all the comments of all rows at once

image


libs/shared/sai-editor/src/sai-editor.service.spec.ts
```diff
@@ -144,5 +144,31 @@ describe('SAI Editor Service', () => {
...
+
+      service.generateComments();
+      tick(100);
+
+      expect(updateDiffQuerySpy).toHaveBeenCalledTimes(1);
+      expect(updateFullQuerySpy).toHaveBeenCalledTimes(1);
+      expect(refreshDatatableSpy).toHaveBeenCalledTimes(1);
+      expect(isRowSelectedSpy).toHaveBeenCalledTimes(1);
+      expect(getNameSpy).toHaveBeenCalledTimes(1);
+      expect(generateCommentSpy).toHaveBeenCalledTimes(1);
```
FrancescoBorzi, line 171: shouldn't we also check on the actual comment that has been added?

libs/shared/sai-editor/src/sai-editor.component.html
```diff
@@ -557,8 +557,15 @@
...
-            Generate comments
+          <button id="generate-comments-btn" class="btn btn-primary btn-sm float-end" (click)="editorService.generateComments(true)">
+            Generate comments (all)
+          </button>
+          <button
+            id="generate-comments-btn-single"
+            class="btn btn-primary btn-sm float-end me-2"
+            (click)="editorService.generateComments()"
+          >
+            Generate comment
```
FrancescoBorzi, line 568: shouldn't we use translations? also the text should be more specific, like "Generate comment for selected row"

FrancescoBorzi: also this button should be disabled if no row is selected

libs/shared/sai-editor/src/sai-editor.service.ts
```diff
@@ -93,22 +93,35 @@ export class SaiEditorService extends MultiRowComplexKeyEditorService<SmartScrip
...
   }
+
+  private async generateSingleComment(row: SmartScripts): Promise<string> {
+    return this.saiCommentGeneratorService.generateComment(
+      structuredClone(this._newRows),
+      { ...row },
+      /* istanbul ignore next */
+      await this.handlerService.getName()?.toPromise(),
+    );
+  }
```
FrancescoBorzi, line 126: missing integration tests

libs/shared/sai-editor/src/sai-editor.service.spec.ts
Helias: everything is mocked, so the comment is undefined here

FrancescoBorzi (approved):

PR #2902, opened 2024-04-30
refactor: remove ngclass

Helias:


libs/features/creature/src/creature-template/creature-template.component.html
```diff
@@ -548,7 +548,7 @@
         </div>
       </div> -->
       </form>
-      <div class="preview-container" [ngClass]="showItemPreview ? 'show-preview' : 'hide-preview'">
+      <div class="preview-container" [class.show-preview]="showItemPreview" [class.hide-preview]="!showItemPreview">
```
FrancescoBorzi, line 551: would this work? `[class]="showItemPreview ? 'show-preview' : 'hide-preview'">`

FrancescoBorzi (approved):

libs/features/creature/src/creature-template/creature-template.component.html
Helias: I think so, but eslint could complain about `double attribute` because `class="something" [class]=".."` it's the same for the strict eslint rules, so I would avoid it

PR #2958, opened 2024-05-15
feat(creature-template): improve selectors, add icons

Helias:


libs/features/creature/src/creature-template/creature-template.component.html
```diff
@@ -203,7 +201,10 @@
             <div class="form-group col-12 col-sm-6 col-md-4 col-lg-3 col-xl-2">
               <label class="control-label" for="RacialLeader">RacialLeader</label>
               <i class="fas fa-info-circle ms-1" placement="auto" [tooltip]="'CREATURE.TEMPLATE.RACIAL_LEADER' | translate"></i>
-              <input [formControlName]="'RacialLeader'" id="RacialLeader" type="number" class="form-control form-control-sm" />
+              <select [formControlName]="'RacialLeader'" id="RacialLeader" class="form-control form-control-sm">
+                <option [ngValue]="0">Disabled</option>
+                <option [ngValue]="1">Enabled</option>
```
FrancescoBorzi, line 206: same

libs/features/creature/src/creature-template/creature-template.integration.spec.ts
```diff
@@ -131,23 +131,42 @@ describe('CreatureTemplate integration tests', () => {
...
+        '`detection_range` = 22, `scale` = 23, `rank` = 1, ' +
+        '`dmgschool` = 1, `DamageModifier` = 26, `BaseAttackTime` = 27, `RangeAttackTime` = 28, `BaseVariance` = 29, ' +
+        '`RangeVariance` = 30, `unit_class` = 1, `unit_flags` = 32, `unit_flags2` = 33, `dynamicflags` = 34, `family` = 1, ' +
+        '`trainer_type` = 1, `trainer_spell` = 37, `trainer_class` = 1, `trainer_race` = 1, `type` = 1, `type_flags` = 41, ' +
         '`lootid` = 42, `pickpocketloot` = 43, `skinloot` = 44, `PetSpellDataId` = 45, `VehicleId` = 46, ' +
-        "`mingold` = 47, `maxgold` = 48, `AIName` = '49', `MovementType` = 50, `HoverHeight` = 51, " +
-        '`HealthModifier` = 52, `ManaModifier` = 53, `ArmorModifier` = 54, `ExperienceModifier` = 55, `RacialLeader` = 56, `movementId` = 57, `RegenHealth` = 58, ' +
+        "`mingold` = 47, `maxgold` = 48, `AIName` = '49', `MovementType` = 2, `HoverHeight` = 51, " +
+        '`HealthModifier` = 52, `ManaModifier` = 53, `ArmorModifier` = 54, `ExperienceModifier` = 55, `RacialLeader` = 1, `movementId` = 57, `RegenHealth` = 0, ' +
```
FrancescoBorzi, line 164: I don't get why is the query changing?

libs/shared/acore-world-model/src/options/creature-icon.ts
```diff
@@ -1,7 +1,7 @@
 import { Option } from '@keira/shared/constants';
 
 export const CREATURE_ICON: Option[] = [
-  { value: '', name: 'None' },
+  { value: null, name: 'None' },
```
FrancescoBorzi, line 4: why?

libs/shared/test-utils/src/lib/editor-page-object.ts
```diff
@@ -46,8 +46,12 @@ export abstract class EditorPageObject<T> extends PageObject<T> {
...
   }
 
-  setSelectValueById(inputId: string, value: number): void {
-    this.setInputValueById(inputId, `${value}: ${value}`);
+  setSelectValueById(inputId: string, value: string | number): void {
+    if (typeof value === 'string') {
+      this.setInputValueById(inputId, `${value}`);
+    } else {
+      this.setInputValueById(inputId, `${value}: ${value}`);
+    }
```
FrancescoBorzi, line 54: not clear why we need this? maybe we should add a comment here with an example

libs/features/creature/src/creature-template/creature-template.integration.spec.ts
Helias: Because the previous values are not available in the current <option> list

libs/shared/acore-world-model/src/options/creature-icon.ts
Helias: Because '' is different than null
If you put '' the query diff will detect a change, with null no

libs/shared/test-utils/src/lib/editor-page-object.ts
Helias: Nice question, I will

libs/shared/constants/src/types/general.ts
```diff
@@ -39,7 +39,7 @@ export interface Flag {
 }
 
 export interface Option {
-  value: number | string;
+  value: number | string | null;
```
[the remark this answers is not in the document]
Helias: I can't put undefined, I need null for this

libs/shared/selectors/src/selectors/boolean-option-selector/boolean-option-selector.component.html
```diff
@@ -0,0 +1,4 @@
+<select [formControl]="control()" [id]="controlName()" class="form-control form-control-sm">
+  <option [ngValue]="0">0 - Disabled</option>
+  <option [ngValue]="1">1 - Enabled</option>
+</select>
```
FrancescoBorzi, line 4: missing tests

libs/shared/selectors/src/selectors/generic-option-selector/generic-option-selector.component.html
```diff
@@ -0,0 +1,5 @@
+<select [formControl]="control()" [id]="controlName()" class="form-control form-control-sm">
+  @for (option of optionList(); track option) {
+    <option [ngValue]="option.value" [label]="option.value + ' - ' + option.name"></option>
+  }
+</select>
```
FrancescoBorzi, line 5: missing tests

libs/shared/selectors/src/selectors/boolean-option-selector/boolean-option-selector.component.html
```diff
@@ -0,0 +1,4 @@
+<select [formControl]="control()" [id]="controlName()" class="form-control form-control-sm">
+  <option [ngValue]="0">0 - Disabled</option>
+  <option [ngValue]="1">1 - Enabled</option>
```
FrancescoBorzi, line 3: what about translations? for "Enabled" and "Disabled"

libs/shared/selectors/src/selectors/boolean-option-selector/boolean-option-selector.component.ts
```diff
@@ -0,0 +1,14 @@
+import { ChangeDetectionStrategy, Component, input } from '@angular/core';
+import { FormControl, FormsModule, ReactiveFormsModule } from '@angular/forms';
+
+@Component({
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  selector: 'keira-boolean-option-selector',
+  templateUrl: './boolean-option-selector.component.html',
+  imports: [FormsModule, ReactiveFormsModule],
```
FrancescoBorzi, line 8: are you sure that just `ReactiveFormsModule` isn't enough?

libs/shared/selectors/src/selectors/boolean-option-selector/boolean-option-selector.component.html
```diff
@@ -0,0 +1,4 @@
+<select [formControl]="control()" [id]="controlName()" class="form-control form-control-sm">
```
FrancescoBorzi, line 1: what do we need `controlName()` for? I think it can be avoided (and the `id` be set in the parent component)

libs/shared/selectors/src/selectors/generic-option-selector/generic-option-selector.component.html
```diff
@@ -0,0 +1,5 @@
+<select [formControl]="control()" [id]="controlName()" class="form-control form-control-sm">
```
FrancescoBorzi, line 1: same as above for the `controlName()`, why do we need this here?

libs/shared/selectors/src/selectors/boolean-option-selector/boolean-option-selector.component.ts
```diff
@@ -0,0 +1,14 @@
...
+  standalone: true,
+})
+export class BooleanOptionSelectorComponent {
+  control = input.required<FormControl>();
+  controlName = input.required<string>();
```
FrancescoBorzi, line 13: why using signals here?

libs/shared/selectors/src/selectors/boolean-option-selector/boolean-option-selector.component.html
FrancescoBorzi: > it's to be coherent with the structure before, so from the id you can access the formControl

I think it is better to move it out, because we do not need it. If you want to access the formControl, you can still do by first accessing the outer component and then querying inside it

PR #3025, opened 2024-07-14
Add Brazilian Portuguese

valberlimabr:
Added the Brazilian Portuguese language from the Portuguese language already existing in the source code, in the Brazilian Portuguese language file, some translations that were missing (not all) were added, as well as improvements were made to the quality of the existing translation, making it more similar to the native Brazilian Portuguese.


FrancescoBorzi:
hi @valberlimabr thanks for your contribution.

Aren't Portuguese and Brazilian Portuguese mutually intelligible languages?

valberlimabr:
hi @FrancescoBorzi 

They are similar languages, but due to the cultural differences between Portugal and Brazil, some words written the same may have different meanings. This is why I chose to create a file for Brazilian Portuguese.

Nyeriah:
The disparities are not meaningful enough to make a difference in this context. Is threre any translation you particularly have issues with? We could instead try to adapt and/or replace wording that may seem dubious.

FrancescoBorzi:
@valberlimabr creating and maintaining 2 different language files seems too much work for such a small benefit. Would you like to edit the Portuguese version instead and make it a bit more fluid?

valberlimabr:
@FrancescoBorzi Thanks for explaining, so I will edit the existing Portuguese language file instead of creating a new file.

TheSCREWEDSoftware:
I know this is closed. but there are so many differences which lack of context can lead to missleading, as those difference meanigful enough to have to have two different PT-PT and PT-BR, i would say no. while the most commun spoken is PT-BR, adding to the existing translation (which i've not looked into, i assume i pt-pt) even if they are mixed between PT-BR and PT-PT it's better than not translation. Even if eventually it becomes either PT-PT or PT-BR.

Nyeriah:
@TheSCREWEDSoftware we will work on those difference individually and try to minimize the disparities rather than add a whole new language. If there are any glaring ones you have in mind we could start from there

FrancescoBorzi:
I agree with Nyeriah. Please @valberlimabr @TheSCREWEDSoftware report any text that is not clear in either of the two languages and we can try to rephrase it so it is clear enough for both.

TheSCREWEDSoftware:
I agree with both of your mesasges now and my comment was related to the 21st of Jully ones. They are many differences that may make a hard to understand but doesnt justfiy a different / new language. While for now is better to join both efforts to have majority of things translated

Issue #3059, opened 2024-10-12
creature_template's content not propely updating after running SQL via the SQL Editor


TheSCREWEDSoftware:
1 - Create (new) creature_template entry

2 - Go to creature_template

Lets use `90000001`

<details>
<summary>Entry Created</summary>

![image](https://github.com/user-attachments/assets/fb7eb35c-220e-4c2e-9baf-4999fa056bf5)

</details>

3 - (you can skip this part if you're using my example entries and SQL) Go back to select creature of your choice, we will use 
Kobold Miner (Entry: 40)

4 - Go that creature's template and click on full-query on top and copy that SQL and edit it

From:
```SQL
DELETE FROM `creature_template` WHERE (`entry` = 40);
INSERT INTO `creature_template` (`entry`, `difficulty_entry_1`, `difficulty_entry_2`, `difficulty_entry_3`, `KillCredit1`, `KillCredit2`, `name`, `subname`, `IconName`, `gossip_menu_id`, `minlevel`, `maxlevel`, `exp`, `faction`, `npcflag`, `speed_walk`, `speed_run`, `speed_swim`, `speed_flight`, `detection_range`, `scale`, `rank`, `dmgschool`, `DamageModifier`, `BaseAttackTime`, `RangeAttackTime`, `BaseVariance`, `RangeVariance`, `unit_class`, `unit_flags`, `unit_flags2`, `dynamicflags`, `family`, `trainer_type`, `trainer_spell`, `trainer_class`, `trainer_race`, `type`, `type_flags`, `lootid`, `pickpocketloot`, `skinloot`, `PetSpellDataId`, `VehicleId`, `mingold`, `maxgold`, `AIName`, `MovementType`, `HoverHeight`, `HealthModifier`, `ManaModifier`, `ArmorModifier`, `ExperienceModifier`, `RacialLeader`, `movementId`, `RegenHealth`, `mechanic_immune_mask`, `spell_school_immune_mask`, `flags_extra`, `ScriptName`, `VerifiedBuild`) VALUES
[...]

PR #3101, opened 2024-10-26
feat(handler-service): implement force reload

Helias:
closes https://github.com/azerothcore/Keira3/issues/3059

### DIFF
--- a/libs/shared/base-abstract-classes/src/service/handlers/handler.service.ts
+++ b/libs/shared/base-abstract-classes/src/service/handlers/handler.service.ts
@@ -52,6 +52,7 @@ export abstract class HandlerService<T extends TableRow> extends SubscriptionHan
       this._selected = `${id}`;
     }
 
+    // To prevent side effects or outdated data, it's essential to reload the entity when the user selects it again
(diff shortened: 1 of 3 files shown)


libs/shared/base-abstract-classes/src/service/handlers/handler.service.ts
```diff
@@ -43,12 +44,18 @@ export abstract class HandlerService<T extends TableRow> extends SubscriptionHan
...
     if (typeof id === 'object') {
       this._selected = JSON.stringify(id);
     } else {
       this._selected = `${id}`;
     }
 
+    if (currentSelected === this._selected) {
+      this.forceReload = true;
+    }
+
```
FrancescoBorzi, line 58: can you please leave a comment explaining why this is needed and when it will be triggered?

FrancescoBorzi (approved):

libs/shared/base-abstract-classes/src/service/handlers/handler.service.ts
Helias: done

Issue #984, opened 2020-11-21
ENTRY ID Errors (Easy Fix)


ghost:
Keira 3 allows you to put extremely large entry IDs for creatures/items such as 18888800003.

This is too big for wow's database to accept the heidiSQL will throw up an error if you try to import the changes.

It could be worth limiting the input of entryIDs to match the maximum size of entryIDs compatible with the database or at least a warning in the info section of ENTRYID.

Field | Type | Null | Key | Default | Extra | Comment
-- | -- | -- | -- | -- | -- | --
entry | mediumint(8) unsigned | NO | PRI | 0

mediumint   | 3 bytes  -8388608 to 8388607                          0 to 16777215

https://www.azerothcore.org/wiki/creature_template#entry

https://stackoverflow.com/questions/2991405/what-is-the-difference-between-tinyint-smallint-mediumint-bigint-and-int-in-m

FrancescoBorzi:
it sounds quite reasonable

PR #3113, opened 2024-10-27
feat(create.component): set max ID value

Helias:
closes https://github.com/azerothcore/Keira3/issues/984

### DIFF
--- a/libs/shared/base-editor-components/src/create/create.component.spec.ts
+++ b/libs/shared/base-editor-components/src/create/create.component.spec.ts
@@ -1,13 +1,13 @@
 import { ComponentFixture, TestBed, waitForAsync } from '@angular/core/testing';
 import { FormsModule } from '@angular/forms';
 import { BrowserModule } from '@angular/platform-browser';
+import { MockHandlerService } from '@keira/shared/base-abstract-classes';
+import { TableRow } from '@keira/shared/constants';
+import { MysqlQueryService } from '@keira/shared/db-layer';
 import { PageObject, TranslateTestingModule } from '@keira/shared/test-utils';
 import { of, throwError } from 'rxjs';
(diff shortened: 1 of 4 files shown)


libs/shared/base-editor-components/src/create/create.component.ts
```diff
@@ -74,6 +76,14 @@ export class CreateComponent<T extends TableRow> extends SubscriptionHandler imp
     );
   }
 
+  protected checkMaxValue(): void {
+    console.log('this.idModel', this.idModel);
+    if (this.idModel > MAX_INT_UNSIGNED_VALUE) {
+      this.idModel = MAX_INT_UNSIGNED_VALUE;
+      console.log('this.idModel # 2', this.idModel);
```
FrancescoBorzi, line 83:
```suggestion
    if (this.idModel > MAX_INT_UNSIGNED_VALUE) {
      this.idModel = MAX_INT_UNSIGNED_VALUE;
```

libs/shared/base-editor-components/src/create/create.component.spec.ts
```diff
@@ -107,6 +109,15 @@ describe('CreateComponent', () => {
     expect(selectSpy).toHaveBeenCalledWith(true, id);
   });
 
+  it('does not allow a higher value than max value', () => {
+    const unallowedIdValue = MAX_INT_UNSIGNED_VALUE + 1;
+    component.idModel = unallowedIdValue;
+
+    component['checkMaxValue']();
```
FrancescoBorzi, line 116: let's not do this unless strictly necessary. Isn't is possible to trigger this from the UI?

Helias: I tried but it was not feasible

Issue #3142, opened 2024-11-06
Bitmask for socketColor_# not displaying correctly


blinkysc:
When using the ... for bitmask values for gems the bitmask applying incorrect values.

To Reproduce:

item 50735
Color_1 = 2 Red
Color_2 = 8 Blue
Color_3 = 4 Yellow
Use ... helper and notice it thinks 2 is bit 1 (Meta) and setting as red will apply 4 as value

![image](https://github.com/user-attachments/assets/a264c21b-9586-4d04-911c-c83c30ac429e)

TheSCREWEDSoftware:
@Helias and @FrancescoBorzi provide me with some insight, i was looking into this issue but i dont see anywhere in the item_template in AC or even in TC where it mentions that ScoketColours should be a bitmask. They are just a value.

You either choose it's a Meta, red, yellow or blue gem. isn't npcflags. right?

AC
![image](https://github.com/user-attachments/assets/92e2d669-6deb-4316-879e-54088bbbecae)

TC
![image](https://github.com/user-attachments/assets/9c9ce5c4-7089-48cc-9c10-45357f6b8c78)

If i introduce the value manually (not the bitmask)

I can change to the proper gem

Helias:
Yes we could switch this into a `keira-generic-option-selector`

FrancescoBorzi:
>  i was looking into this issue but i dont see anywhere in the item_template in AC or even in TC where it mentions that ScoketColours should be a bitmask

Eeemh sorry to be the party pooper but it actually really looks like it is a bitmask. Because 1, 2, 4, 8 are indeed powers of 2

And there are indeed composite values (even rare):

![image](https://github.com/user-attachments/assets/bc02ebaa-0f58-46e3-8d53-ab48070938de)

14 for example is 8+4+2 so it must be bitmask. The problem is probably that the values are a bit off. I'm going to look into it.

FrancescoBorzi:
I'm thinking that the fix should rather be:

![image](https://github.com/user-attachments/assets/69f5df72-ceea-4677-895d-74943dbf3403)

As the `bit` in flags represent the POWER of 2 (not the final value).

In this example:

- bit `0` means `2^0` = `1`
- bit `1` means `2^1` = `2`
- bit `2` means `2^2` = `4`
- bit `3` means `2^3` = `8`

FrancescoBorzi:
[...]

PR #3149, opened 2024-11-10
Updated the Socket (Gem) Colour - Selector "keira-flags-selector-btn" => "generic-option-selector"

TheSCREWEDSoftware:
Fixes the issue: https://github.com/azerothcore/Keira3/issues/3142

Removed: `libs\shared\acore-world-model\src\flags\socket-color.ts` as was no longer needed. 
Created: `libs\shared\acore-world-model\src\options\socket-color.ts` to work with the new option instead of flag and with the values needed to dispaly.
Updated: `libs\shared\acore-world-model\src\index.ts` removed the flag for socket-colour and added the options one.
Updated: `libs\features\item\src\item-template\item-template.component.ts` just the import for `GenericOptionSelectorComponent` (kept the socket import)
Updated: `libs\features\item\src\item-template\item-template.component.html` Change selector from keira-flags-selector-btn" => "generic-option-selector"

<details>
<summary>Images of the before and after</summary>

Before
image

Now
image

</details>

Note: The file name and commit have not the proper name or description due to github desktop telling me to take a walk because of husky. New branched created (and tested again keira, working fine)


FrancescoBorzi:
Thank you very much for contributing. However, wait a second before merging this, see https://github.com/azerothcore/Keira3/issues/3142#issuecomment-2466773501

FrancescoBorzi:
@TheSCREWEDSoftware I truly appreciate you trying to fix this issue and our bad if we haven't communited it earlier, but I believe this should stay a flag. The issue is the value of the bits that must be updated: https://github.com/azerothcore/Keira3/pull/3150

If you think I'm mistaken, feel free to comment and we can reopen this :)

PR #3150, opened 2024-11-10
fix(item_template): fix socket colors bit values

FrancescoBorzi:
- Closes https://github.com/azerothcore/Keira3/issues/3142 https://github.com/azerothcore/Keira3/pull/3149

image

image

image

image

image

### DIFF
--- a/libs/shared/acore-world-model/src/flags/socket-color.ts
+++ b/libs/shared/acore-world-model/src/flags/socket-color.ts
@@ -1,8 +1,8 @@
 import { Flag } from '@keira/shared/constants';
 
 export const SOCKET_COLOR: Flag[] = [
-  { bit: 1, name: 'Meta' },
-  { bit: 2, name: 'Red' },
-  { bit: 4, name: 'Yellow' },
-  { bit: 8, name: 'Blue' },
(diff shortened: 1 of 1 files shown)


TheSCREWEDSoftware:
My question still stands related to the bitmask. I understand they are only 4 results. But this value 1 per slot. You can have 3 slots per item. And each slot has 4 options. Meta, Red, Yellow or Blue, you can't mix them. 

And even if this value was to "calculate" for the intermedie values (purple, orange, green, for the bonus socket value) it would only count (if not wrong) how many of X colour are present

TheSCREWEDSoftware:
Take a look here https://wowdev.wiki/ItemCache.wdb when has a bitmask value it's defined as such, otherwise is a interger in the colour socket it's an interger

FrancescoBorzi:
> My question still stands related to the bitmask. I understand there are only 4 results (with bitmask above these 4 values

@TheSCREWEDSoftware I would say let's first merge this PR as it is at least correcting the value and solving the main bug for the users. Then we can discuss elsewhere whether it makes sense to keep this a flag or not.

Helias (approved):

Issue #3131, opened 2024-11-03
[Enhancement] - SQL Editor - Syntax Colouring!


TheSCREWEDSoftware:
![image](https://github.com/user-attachments/assets/00f81259-d100-4345-a5e5-353a90cf6d6e)

```sql
SELECT `entry`, `name`, `subname`, `minlevel`, `maxlevel`, `AIName`, `ScriptName` 
FROM `creature_template` 
WHERE `entry` > 100 
ORDER BY `entry` ASC 
LIMIT 100
```

Edit: Also removing the auto-correction red lines from that textbox

Note: Also placing the Link for SQL Language Tutorial on the top right

![image](https://github.com/user-attachments/assets/38b3eaf0-6765-41b0-8947-dd4e0445c5e3)

PR #3173, opened 2024-11-20
feat(sql-editor-minor-improvements)

alecsci:
Progresses: [3131](https://github.com/azerothcore/Keira3/issues/3131)
- disabled spellcheck
- moved the sql on right top side

### DIFF
--- a/libs/features/sql-editor/src/sql-editor.component.html
+++ b/libs/features/sql-editor/src/sql-editor.component.html
@@ -1,11 +1,19 @@
 <div class="container-fluid">
   <div class="content-block">
-    <p class="lead mt-3 mb-2">
-      {{ 'SQL_EDITOR.TITLE' | translate }}
-      <i class="fas fa-info-circle ms-1" placement="auto" [tooltip]="'SQL_EDITOR.TOOLTIP' | translate"></i>
-    </p>
+    <div class="row">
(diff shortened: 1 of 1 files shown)


alecsci:
Hi @FrancescoBorzi, @Helias 

For the syntax coloring, I would need some guidance.

I found [highlight.js](https://highlightjs.org/), but not sure if it's integrating correctly with the existing component lifecycle.
I need to register which language is supported, I suppose in the constructor of the sql-component-editor.ts ?

alecsci:
I got it now, they are already using ngx-highlightjs but in other parts of the application.

FrancescoBorzi (approved):

FrancescoBorzi:
@alecsci sorry it took a while to review this PR but I'm extremely busy lately. The code looks good. I was wondering if this is already some improvement we could merge or to be considered yet a Work In Progress (I haven't tested it locally, just checked the code).

If you tested this already and it already makes it a bit better, I'd just merge it.

alecsci:
Hi @FrancescoBorzi 
Yes, I tested it only locally and works, it doesn't completely finish the requirements of the issue, it progresses most of it.

The only missing part to be completely done is the sql code ngx highlighting in the editor menu which I couldn't find yet a solution.

PR #3174, opened 2024-11-20
fix: display only the error message for issue #3132

### LINKED ISSUE #3132: [Enhancement] - Simplification of SQL Errors
Error here is `entr` instead of `entry`
 
 Note: Maybe a link for the default browser with default search engine just copy and pastes the error in the browser like HeidiSQL does [here](https://www.ecosia.org/search?q=SQL%20Error%20(1054):%20Unknown%20column%20%27entr%27%20in%20%27where%20clause%27uery)

```sql
SELECT `entry`, `name`, `subname`, `minlevel`, `maxlevel`, `AIName`, `ScriptName` 
FROM `creature_template` 
WHERE `entr` > 100 
ORDER BY `entry` ASC 
LIMIT 100
```
## Keira
image

## Heidi SQL and Windows CMD

image

image

alecsci:
fix #3132

### DIFF
--- a/libs/shared/base-editor-components/src/query-output/query-error/query-error.component.html
+++ b/libs/shared/base-editor-components/src/query-output/query-error/query-error.component.html
@@ -3,7 +3,7 @@
     <table>
       <tr>
         <td>Error:</td>
-        <td>{{ error.stack }}</td>
+        <td>{{ error.message }}</td>
       </tr>
       <tr>
         <td>Code:</td>


alecsci (commented):
For the users this is better, for the devs this is worse because we lose the stack-trace, so maybe we should have at the end of the alert message a field for the stack-trace ?

alecsci:
Not sure if it's worth doing that much implementation to hide some technical info. 
I would like to keep the error component dumb as possible.

We could show both information, but put first the error message, and then the last field can be the stack trace.
@Francesco-Borzi wdyt?

Helias:
We could show both but in a separate way, so the user can focus on the error message first and then if it's useful can also read the stack.

PR #3194, opened 2024-12-15
Add new creature formations component

### LINKED ISSUE #3060: add creature_formations editor under creature

Exitare:
closes #3060

TODOS:

- [ ] Integration tests need to be updated.

I receive this error when trying to navigate to the new component:
image

New sidebar:
image

The component is therefore not loading. When wrapping this piece of code in a try catch:
```
protected disableEntityIdField(): void {
    this._form.controls[this._entityIdField as string].disable();
  }
```
located in editor.service.ts (L80-82)

the page is being displayed. 

I have not figured out what I am missing, or whats wrong with it. 

@FrancescoBorzi @Helias your insights & feedback are greatly appreciated.


Exitare:
Thanks for pointing me in the right direction regarding the issue :)

Exitare:
@Helias 
The problem however is that the columns are spelled GUID in the database all caps. 
Without the upper case letter, the column names are not read correctly. Maybe thats something we have to change in the AC database to make it work with Keira ? Or is there a way to get around that?

Correctly spelled:
image

image

Incorrectly spelled:
image


it also causes issues with the sql generation as the bindings do not work anymore:
image

Helias:
Hey, nono, it's not necessary to change AC, Keira3 must adapt to AC, I can change the column's name and make it work

Helias:
renamed, you can update your sources, now it works with the AC columns names

Exitare:
Thanks for your help here. 
However, I am confused. Looking at your changes, I cannot find where I went wrong from the beginning. Everything was named correctly and yet the page didn't open because of the issue I encountered which can be seen in the image above. 
You changed the naming and everything worked "suddenly". 

What exactly was the problem here?

Yes, all unit tests are failing. I wanted to get the page up and running before fixing the tests.

Exitare:
Tests passing. Ready for review

Helias (approved):

PR #3210, opened 2024-12-20
feat: new acore_string editor

Exitare:
Adds a new editor for the acore_string table for the section "Other texts".
I realized yesterday that there isn't an editor or at least search field for this table and decided to implement it.

image

PR #3212, opened 2024-12-21
Convert editorService from public to protected

Exitare:
As mentioned [here](https://github.com/azerothcore/Keira3/pull/3210#discussion_r1894425492) the editor in components can be made protected. 
This PR does that for all currently available editors. 
I hope that this change is helpful

### DIFF
--- a/libs/shared/sai-editor/src/sai-editor.component.ts
+++ b/libs/shared/sai-editor/src/sai-editor.component.ts
@@ -87,7 +87,7 @@ import { EditorButtonsComponent, QueryOutputComponent } from '@keira/shared/base
   ],
 })
 export class SaiEditorComponent extends MultiRowEditorComponent<SmartScripts> implements OnInit {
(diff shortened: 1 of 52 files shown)


libs/shared/sai-editor/src/sai-editor.component.ts
```diff
@@ -87,7 +87,7 @@ import { EditorButtonsComponent, QueryOutputComponent } from '@keira/shared/base
   ],
 })
 export class SaiEditorComponent extends MultiRowEditorComponent<SmartScripts> implements OnInit {
-  override readonly editorService = inject(SaiEditorService);
+  public override readonly editorService = inject(SaiEditorService);
```
Helias, line 90: we should define a codestyle about writing `public`, I personally don't write the code that is unnecessary, but I see that there is no consistency in Keira3 @FrancescoBorzi

Helias (approved):

PR #3215, opened 2024-12-23
ci: add new workflow to sync i18n files

### LINKED ISSUE #3211: i18n files are not in sync.
While adding new translations I realized that some files have Key Value Pairs that are unique to the file and are not used in other files. 

E.g.

es.json

```json
 "PAGE_TEXT": "Page Text",
 "PAGE_TEXTS": {
        "ID": "The ID of the text in the page. This number is unique to every text ID.",
        "NEXT_PAGE_ID": "The ID of the next page's text ID.",
        "TEXT": "The actual text. The message in this field will be shown as the text on a page."
}
```

While almost all other files use:
```json
"SELECT_PAGE_TEXT": "Select Page Text",
"PAGE_TEXT": "Page Text"
```

I dont know what the best thing to do here is, or if it is even worth the time to correct that, as probably the majority will use Keira in English anyway?

Exitare:
Using en.json as default to sync all keys across the available json files. 
Commits the adjustments if necessary.
Closes #3211

Example fix can be seen here:
https://github.com/Exitare/Keira3/commit/435a29eca18f67351cfd4b7e8edaf47335aefffc

### DIFF
--- a/scripts/sync_i18n.py
+++ b/scripts/sync_i18n.py
@@ -3,7 +3,7 @@ import json
 from pathlib import Path
 
 
-def merge_with_en(en_data, target_data):
+def merge_with_en(en_data: dict, target_data: dict) -> dict:
     """
     Recursively merge en.json keys into target translation file.
     - Adds missing keys from en.json.
@@ -24,7 +24,7 @@ def merge_with_en(en_data, target_data):
     return merged
 
 
-def sync_keys(translations_dir, en_file):
+def sync_keys(translations_dir: Path, en_file: Path):
     # Load en.json as the reference file
     with open(en_file, 'r', encoding='utf-8') as f:
         en_data = json.load(f)
@@ -47,5 +47,5 @@ def sync_keys(translations_dir, en_file):
 
 if __name__ == "__main__":
     translations_dir = Path("apps", "keira", "src", "assets", "i18n")  # Update to your directory
-    en_file = translations_dir / "en.json"
+    en_file = Path(translations_dir, "en.json")
     sync_keys(translations_dir, en_file)
--- a/scripts/sync_i18n.py
+++ b/scripts/sync_i18n.py
@@ -1,0 +1,51 @@
+import os
+import json
+from pathlib import Path
+
+
+def merge_with_en(en_data, target_data):
+    """
+    Recursively merge en.json keys into target translation file.
+    - Adds missing keys from en.json.
+    - Preserves existing keys and values in the target file.
+    - Removes keys not present in en.json.
+    """
+    merged = {}
+    for key, value in en_data.items():
+        if key in target_data:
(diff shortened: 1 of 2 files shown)


scripts/sync_i18n.py
```diff
@@ -0,0 +1,51 @@
+import os
+import json
+from pathlib import Path
+
+
+def merge_with_en(en_data, target_data):
```
Helias, line 6: I recommend adding type hints to the variables.
```python
def merge_with_en(en_data: dict, target_data: dict) -> dict:
```

scripts/sync_i18n.py
```diff
@@ -0,0 +1,51 @@
...
+                merged[key] = merge_with_en(value, target_data[key])
+            else:
+                merged[key] = target_data[key]  # Preserve existing value
+        else:
+            merged[key] = value  # Add missing key with value from en.json
+
+    return merged
+
+
+def sync_keys(translations_dir, en_file):
```
Helias, line 27: sam here

Helias:
I like a lot this PR 🚀

Exitare:
Done ;)

scripts/sync_i18n.py
```diff
@@ -0,0 +1,51 @@
...
+
+            # Save the synchronized file
+            with open(file_path, 'w', encoding='utf-8') as f:
+                json.dump(synced_data, f, ensure_ascii=False, indent=2)
+
+            print(f"Synced {file}")
+
+
+if __name__ == "__main__":
+    translations_dir = Path("apps", "keira", "src", "assets", "i18n")  # Update to your directory
```
Helias, line 49:
```suggestion
    translations_dir = Path("apps", "keira", "src", "assets", "i18n")
```

scripts/sync_i18n.py
```diff
@@ -0,0 +1,51 @@
...
+                merged[key] = merge_with_en(value, target_data[key])
+            else:
+                merged[key] = target_data[key]  # Preserve existing value
+        else:
+            merged[key] = value  # Add missing key with value from en.json
+
+    return merged
+
+
+def sync_keys(translations_dir: Path, en_file: Path):
```
Helias, line 27:
```suggestion
def sync_keys(translations_dir: Path, en_file: Path) -> None:
```

Helias (approved):

Helias:
Thanks a lot!

FrancescoBorzi:
@Exitare @Helias the build is failing on master branch

Exitare:
@FrancescoBorzi  because it's not allowed to push on the protected branch. What's the preferred solution here?

FrancescoBorzi:
@Exitare the preferred solution would be having it in the branch instead of on main, so we could disable this check on main

PR #3233, opened 2025-01-07
Update editor to jump back to first page after a new search was triggered

### LINKED ISSUE #3232: [Enchantement] - While searching for any data if you're in any page but 1 take you back to page 1
The easiest example I can give here. Search (in creature) for `Lich King` and go to page `2` now search for `onyxia`. It should take you back to page `1` (as it does when you go back from selecting a creature). It takes you to page `2`

Exitare:
Closes #3232

After a click or pressing return the page will jump back to Page 1.
Created a new interface for type safety.

### DIFF
--- a/libs/features/creature/src/select-creature/select-creature.component.html
+++ b/libs/features/creature/src/select-creature/select-creature.component.html
@@ -41,6 +41,7 @@
       <keira-highlightjs-wrapper [code]="selectService.query" />
     </form>
 
+    {{ selectService.pageOffset }}
     @if (selectService.rows) {
       <div>
         <ngx-datatable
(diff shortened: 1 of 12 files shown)


libs/features/creature/src/select-creature/select-creature.component.html
```diff
@@ -41,6 +41,7 @@
       <keira-highlightjs-wrapper [code]="selectService.query" />
     </form>
 
+    {{ selectService.pageOffset }}
```
Helias, line 44: do we want to show this?

Exitare: Haha, damn. Good catch. No we don't need that ^^

PR #3235, opened 2025-01-08
chore(creature_template): update initial unit class from being 0 to 1 (warrior class)

### LINKED ISSUE #3230: [Bug] - Upon creating a new "creature_template", unit_class is blank.
While keira allows you to not choose any `unit_class` this will default to 0. Which the core automatically will set 0 to the next aviaible value which is 1 (i think is what i understood by the message). So it should at least in keira upon creating a new creature show the `1 - warrior` class. this is the rror you get `has invalid unit_class (0) in creature_template. Set to 1 (UNIT_CLASS_WARRIOR)`

Exitare:
Closes #3230


Helias:
thank you!

PR #3242, opened 2025-01-11
Created pull_request_template.md

TheSCREWEDSoftware:
Addes a template to the PRs (intially i thought only having "Closes" tell me if the rests is unsucessary. 

Also i'm not sure how `gitignore` this file (in the case if accepeted)


Helias:
Hi, thanks for your contribution but I have mix feelings about this, I would remove any mandatory part or PR template here to make the development faster
The Keira3 PRs are not so frequent and who works on Keira3 does not need to be guided I think

If we will really need some structured PR we will consider again these options

PR #3250, opened 2025-01-13
Improve item quality visibility

### LINKED ISSUE #3247: [Enchantement] - Item Quality easier to see.
They are 3 ways for this to be implemented.

1 - All items have their name colours based on their quality (as item_template) shows but not any other place. Like selectors, loots outside of item_template. Also this could be applies to the "Quality Selector".

2 - Same above but instead of names a colour around the border of the icon.

3 - A preview of the item.

Also, bright colours of qualities don't go well with the white theme of keira.

Exitare:
Closes #3247

I adjusted the handler to hold a new property, for the item quality. Maybe there is a better word for it?
Everything works, but two tests fail and I cannot figure out why.
image
Thats my output. 
Any ideas @Francesco-Borzi or @Helias 

Maybe you have another idea on how to solve that?

### DIFF
--- a/libs/shared/base-editor-components/src/top-bar/top-bar.component.ts
+++ b/libs/shared/base-editor-components/src/top-bar/top-bar.component.ts
@@ -10,4 +10,12 @@ export class TopBarComponent {
   @Input({ required: true }) selected: string | undefined;
   @Input({ required: true }) selectedName: string | undefined;
   @Input({ required: true }) isNew = false;
+  @Input({ required: false }) customClass: string | undefined = undefined;
+
+  get scssClass() {
+    if (this.customClass) {
+      return 'main-text text-truncate ' + this.customClass;
+    }
+    return 'main-text text-truncate';
+  }
 }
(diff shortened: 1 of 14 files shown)


libs/shared/base-editor-components/src/top-bar/top-bar.component.ts
```diff
@@ -10,4 +10,12 @@ export class TopBarComponent {
   @Input({ required: true }) selected: string | undefined;
   @Input({ required: true }) selectedName: string | undefined;
   @Input({ required: true }) isNew = false;
+  @Input({ required: false }) customClass: string | undefined = undefined;
+
+  get scssClass() {
```
Helias, line 15:
```suggestion
  get scssClass(): string {
```

Helias:
The problem seems to be that the code coverage does not reach the 100% threshold, so we should add some new unit-test to cover all lines of code

Exitare:
Thanks. I thought about that, but also had in mind that there isn't 100% coverage anyway. I will work on it. 
Again thanks for pointing me in this direction :)

Exitare:
@Helias  done :)

Helias (approved):

Helias:
Merged! ✅

PR #3252, opened 2025-01-14
Add default behaviour override for flag selectors.

### LINKED ISSUE #3239: [Bug] - Item_template AllowableClass and Race are missing "-1" for all.
Bitmask currently doesn't (visually) and option support -1 (if you manually type it it will work)

image

image
(since doesnt know about -1 doesnt update properly)

Exitare:
Closes #3239

It's possible to override the default behavior for the flag selector now. If the behavior is overridden, it defaults to -1 when all bits are unset. Otherwise everything stays the same.

### DIFF
--- a/libs/shared/selectors/src/selectors/flags-selector/flags.service.ts
+++ b/libs/shared/selectors/src/selectors/flags-selector/flags.service.ts
@@ -7,7 +7,9 @@ import { Flag } from '@keira/shared/constants';
 export class FlagsService {
   getBitsFromValue(value: number, bitsCount: number): boolean[] {
     const bits = new Array<boolean>(bitsCount);
-    const binaryStr: string = value.toString(2).split('').reverse().join('');
+
+    let binaryStr: string = '0'.repeat(bitsCount);
+    if (value !== -1) binaryStr = value.toString(2).split('').reverse().join('');
 
     for (let i = 0; i < bitsCount; i++) {
       bits[i] = parseInt(binaryStr[i], 10) === 1;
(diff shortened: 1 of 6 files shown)


Exitare:
@Helias anything missing here ?

libs/shared/selectors/src/selectors/flags-selector/flags.service.ts
```diff
@@ -7,7 +7,9 @@ import { Flag } from '@keira/shared/constants';
 export class FlagsService {
   getBitsFromValue(value: number, bitsCount: number): boolean[] {
     const bits = new Array<boolean>(bitsCount);
-    const binaryStr: string = value.toString(2).split('').reverse().join('');
+
+    let binaryStr: string = '0'.repeat(bitsCount);
+    if (value !== -1) binaryStr = value.toString(2).split('').reverse().join('');
```
FrancescoBorzi, line 12: does this 1-line work too?

```suggestion
    const binaryStr: string = value !== -1 ? value.toString(2).split('').reverse().join('') : '0'.repeat(bitsCount);
```

libs/shared/selectors/src/selectors/flags-selector/flags.service.ts
```diff
@@ -16,7 +18,10 @@ export class FlagsService {
     return bits;
   }
 
-  getValueFromBits(bits: boolean[]): number {
+  getValueFromBits(bits: boolean[], overrideDefaultBehavior: boolean = false): number {
+    // override default behavior if all bits are false
+    if (overrideDefaultBehavior && bits.every((bit) => !bit)) return -1;
```
FrancescoBorzi, line 23:
```suggestion
    if (overrideDefaultBehavior && bits.every((bit) => !bit)) { 
      return -1;
    }
```

FrancescoBorzi (approved):

FrancescoBorzi:
Just a couple of minor suggestions and approved already!

Helias (approved):

Helias:
Thank you a lot!

PR #3256, opened 2025-01-18
Add logic to allow warnings for input fields

### LINKED ISSUE #3236: [Enchantement] - Mandotory fields colour/warning.
Currently any mandatory fields are filled with default data. And if you do try to run SQL it will give you an error. (as it should).

But we while a mandatory field is empty (while selected) or was empty (upon un-selecting or selecting something else) they could have a red border or any feedback that the field can't be empty. Alternative but it's uglier it just adding the "*" like forms do.

Exitare:
Does address but does not close #3236.
Tests are failing, because I dont want to fix them before this idea is either accepted or denied. 
Right now, this is only working for the creature template max level field.

image

Feedback is welcome

## Summary by CodeRabbit

- **New Features**
  - Added inline validation feedback for invalid, touched form fields, including required-field and generic error messages.
  - Added required-field validation across editor forms.
  - Added a Models information section and direct link to the Creature Template Model editor.
  - Improved form validation tracking to reflect the overall validity of edited data.

- **Bug Fixes**
  - Improved form updates after reloads and handling of missing or invalid fields.


FrancescoBorzi:
Build failing

Exitare:
Works on my end. Can you please provide more information ?

Helias:
We have a lint rule that notifies if a component is not using the OnPush strategy, it's complaining about it
image

This is the issue reported by the pipeline

Exitare:
Thank you, but the failed test is intentional.
I would love to hear some feedback on my proposed solution before I fix any tests that will break due to the change. If the solution is rejected I didn't waste too much time :) 
Hope that makes sense!

Helias:
ok, I will test this PR locally and let you know ;-)

Exitare:
Thank you :)

Exitare:
I also would like to mention, that this will **NOT** address any editors. The example I added for the maxLevel in creature template is just for showcasing purposes of the proposed solution. The PR aims to provide an easy framework to add form validation, not to add form validation everywhere it is needed. This needs to happen gradually, otherwise this will be a gigantic PR.

Helias:
I don't dislike the approach but I am wondering if can be done only with:
```html
              <input
                keiraInputValidation
                [formControlName]="'maxlevel'"
                id="maxlevel"
                type="number"
                class="form-control form-control-sm"
              />
```

Without requiring any other information or adding components, because may you can find a way to get the formcontrol from the input and you could spawn dynamically a component next to the <input> without already writing a placeholder component `<keira-validation-feedback>`.
If this is not feasible without using the `<keira-validation-feedback>` component and the input parameter `[keiraInputValidation]="editorService.form.get('maxlevel')"` is mandatory, then I can approve this approach.

Thanks by the way for implementing this proof of concept of warnings! it's a good job

Exitare:
Thank your for providing feedback. I will try to make it work. I also had a solution in mind that minimizes the changes required to all editors as this can introduce quite the overhead in work. I will explore possible solutions and hopefully something will work :)

Exitare:
In theory I could do something like this:

`import { Directive, ViewContainerRef } from '@angular/core';

@Directive({
  selector: '[dynamicComponentHost]',
})
export class DynamicComponentHostDirective {
  constructor(public viewContainerRef: ViewContainerRef) {}
}`

`import { Component, ComponentFactoryResolver, OnInit } from '@angular/core';
import { KeiraValidationFeedbackComponent } from './keira-validation-feedback.component';
import { DynamicComponentHostDirective } from './dynamic-component-host.directive';

@Component({
  selector: 'app-dynamic-form',
  template: `
    <div class="form-group col-12 col-sm-6 col-md-4 col-lg-3 col-xl-2">
      <label for="maxLevel">Max Level</label>
      <input
        [keiraInputValidation]="editorService.form.get('maxlevel')"
        [formControlName]="'maxlevel'"
        id="maxlevel"
        type="number"
        class="form-control form-control-sm"
        (blur)="showValidationFeedback()"
      />
      <ng-template dynamicComponentHost></ng-template>
    </div>
  `,
})
export class DynamicFormComponent implements OnInit {
  @ViewChild(DynamicComponentHostDirective, { static: true })
  dynamicHost!: DynamicComponentHostDirective;

  constructor(
    private resolver: ComponentFactoryResolver,
    public editorService: EditorService
  ) {}

  ngOnInit(): void {}

  showValidationFeedback(): void {
    const control = this.editorService.form.get('maxlevel');
    if (control?.invalid && control.touched) {
      const viewContainerRef = this.dynamicHost.viewContainerRef;
      viewContainerRef.clear(); // Clear any existing components
      const componentFactory =
        this.resolver.resolveComponentFactory(KeiraValidationFeedbackComponent);
      const componentRef = viewContainerRef.createComponent(componentFactory);
      componentRef.instance.control = control; // Pass the control to the component
    }
  }
}`

Used like this:
`<div class="form-group col-12 col-sm-6 col-md-4 col-lg-3 col-xl-2">
  <label for="maxLevel">Max Level</label>
  <input
    [keiraInputValidation]="editorService.form.get('maxlevel')"
    [formControlName]="'maxlevel'"
    id="maxlevel"
    type="number"
    class="form-control form-control-sm"
    (blur)="showValidationFeedback()"
  />
  <ng-template dynamicComponentHost></ng-template>
</div>`


But personally, I think this is a lot of boilerplate code and super convoluted too, for a worse outcome in terms of readability and most likely maintainability too.

If you have a better solution or something I should check out, please let me know :)

Helias:
I was convinced that everything can be implemented inside the directive attribute making it agnostic and generic for all inputs without requiring any input information.

```typescript
import { Directive, ElementRef, inject, OnInit, Renderer2 } from '@angular/core';
import { AbstractControl, NgControl } from '@angular/forms';
import { SubscriptionHandler } from '@keira/shared/utils';

@Directive({
  selector: '[keiraInputValidation]',
  standalone: true,
})
export class InputValidationDirective extends SubscriptionHandler implements OnInit {
  private readonly el: ElementRef = inject(ElementRef);
  private readonly renderer: Renderer2 = inject(Renderer2);
  private readonly ngControl: NgControl = inject(NgControl);

  private errorDiv: HTMLElement | null = null;

  ngOnInit(): void {
    const control = this.ngControl.control;

    if (!control) {
      return;
    }

    this.subscriptions.push(
      control.statusChanges?.subscribe(() => {
        this.updateErrorMessage(control);
      }),
    );
  }

  private updateErrorMessage(control: AbstractControl): void {
    if (this.errorDiv) {
      this.renderer.removeChild(this.el.nativeElement.parentNode, this.errorDiv);
      this.errorDiv = null;
    }

    if (control?.touched && control?.invalid) {
      this.errorDiv = this.renderer.createElement('div');
      this.renderer.addClass(this.errorDiv, 'error-message');

      const errorMessage = control?.errors?.['required'] ? 'This field is required' : 'Invalid field';

      const text = this.renderer.createText(errorMessage);
      this.renderer.appendChild(this.errorDiv, text);

      const parent = this.el.nativeElement.parentNode;
      this.renderer.appendChild(parent, this.errorDiv);
    }
  }
}
``` 

```html
            <div class="form-group col-12 col-sm-6 col-md-4 col-lg-3 col-xl-2">
              <label for="maxLevel">Max Level</label>
              <input keiraInputValidation [formControlName]="'maxlevel'" id="maxlevel" type="number" class="form-control form-control-sm" />
            </div>
```

This will allow us to add the directive `keiraInputValidation` to any input without further boilerplate or implementation, moreover, now the directive will be self-contained and it does not require any other helper component like the keira3-validation-feedback.
I am not a fan of the `Renderer2` service but if we need to manipulate the DOM it's the solution for Angular.

You can find the commit here https://github.com/azerothcore/Keira3/commit/7021a3a5f677c21d962817227c202b9ea517fe6a
The branch name is [warnings-directive](https://github.com/azerothcore/Keira3/compare/warnings-directive)

Exitare:
@Helias 
Wow! thank you. I wasn't aware of this possibility. That looks great. I will copy it over :)

Exitare:
@Helias 
I added the directive and it works, with a twist. 
If the creature template is opened up the first time one can clear the field and put another field in focus (e.g. click on minlevel), the error message will not show up. One has to enter a value again and then delete it, to actually trigger the error message. 
I will look into this. 

Additionally, if we add the validation, we are just one step shy of adding a subject that the keira edit button can subscribe to, to disable them if validation fails. 
What are your thoughts on that?

Exitare:
I added a workaround, by adding markedAsTocuched() when the field is invalid. But maybe thats not as good as a solution as it can be?

Helias:
> I added a workaround, by adding markedAsTocuched() when the field is invalid. But maybe thats not as good as a solution as it can be?

it could be in this case, let's keep it

libs/shared/directives/src/validate-input.directive.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  ngOnInit(): void {
+    const control = this.ngControl.control;
+
+    if (!control) {
+      return;
+    }
+
+    this.subscriptions.push(
+      control.statusChanges?.subscribe(() => {
```
Helias, line 26: reading this, we could put a debounceTime here and/or a distinctUntilChanged in a pipe

Exitare: Done

libs/shared/directives/src/validate-input.directive.ts
```diff
@@ -0,0 +1,58 @@
...
+        const errorMessage = control.errors?.['required'] ? 'This field is required' : 'Invalid field';
+
+        const text = this.renderer.createText(errorMessage);
+        this.renderer.appendChild(this.errorDiv, text);
+
+        this.renderer.appendChild(parent, this.errorDiv);
+      }
+    }
+
+    this.validationService.validationPassed$.next(!!control?.valid);
```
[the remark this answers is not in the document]
Exitare: That is indeed true. Let me fix that.

libs/shared/base-abstract-classes/src/service/editors/editor.service.ts
```diff
@@ -82,10 +82,19 @@ export abstract class EditorService<T extends TableRow> extends SubscriptionHand
   }
 
   protected initForm(): void {
+    // Create an instance of the entity class to access default values
+    const defaultValues = new this._entityClass();
+
+    // Initialize the FormGroup
```
Helias, line 88: this comment seems a bit useless, the code is readable enough

libs/shared/base-abstract-classes/src/service/editors/editor.service.ts
```diff
@@ -82,10 +82,19 @@ export abstract class EditorService<T extends TableRow> extends SubscriptionHand
...
     this._form = new FormGroup<ModelForm<T>>({} as any);
 
+    // Loop through the fields and initialize controls with default values
     for (const field of this.fields) {
-      this._form.addControl(field, new FormControl());
+      const defaultValue = defaultValues[field]; // Get the default value dynamically
```
Helias, line 93: same here for the comment and the `use the default value` below

libs/shared/base-abstract-classes/src/service/editors/single-row-editor.service.spec.ts
```diff
@@ -14,7 +12,7 @@ describe('SingleRowEditorService', () => {
 
   beforeEach(() =>
     TestBed.configureTestingModule({
-      imports: [RouterTestingModule],
+      imports: [],
```
Helias, line 15: you could remove entirely this line
```suggestion
```

libs/shared/base-abstract-classes/src/service/editors/single-row-editor.service.spec.ts
```diff
@@ -162,4 +160,77 @@ describe('SingleRowEditorService', () => {
...
+        writable: true,
+      });
+    });
+
+    it('should set values for valid fields in the form', () => {
+      service['updateFormAfterReload']();
+
+      expect(mockForm.controls.id.setValue).toHaveBeenCalledWith(123); // Valid field
+      expect(mockForm.controls.name.setValue).toHaveBeenCalledWith('Test Name'); // Valid field
+      expect(mockForm.controls.guid.setValue).toHaveBeenCalledWith(456); // Valid field
```
Helias, line 199: I would reduce the redundant comments here `// Valid field`

libs/shared/base-abstract-classes/src/service/editors/single-row-editor.service.ts
```diff
@@ -17,6 +17,8 @@ export abstract class SingleRowEditorService<T extends TableRow> extends EditorS
     protected override handlerService: HandlerService<T>,
   ) {
     super(_entityClass, _entityTable, _entityIdField, handlerService);
+    console.log('Entity Class');
+    console.log(_entityClass);
```
Helias, line 21:
```suggestion
```

libs/shared/base-editor-components/src/query-output/query-output.component.ts
```diff
@@ -21,18 +21,28 @@ import { QueryErrorComponent } from './query-error/query-error.component';
...
   private readonly modalService = inject(BsModalService);
+  private readonly validationService = inject(ValidationService);
 
   @Input() docUrl!: string;
   @Input() editorService!: EditorService<T>;
   @Output() executeQuery = new EventEmitter<string>();
   selectedQuery: 'diff' | 'full' = 'diff';
   private modalRef!: BsModalRef;
-
+  protected validationPassed: boolean = true;
```
Helias, line 34:
```suggestion
  protected validationPassed = true;
```

libs/shared/base-editor-components/src/query-output/query-output.component.ts
```diff
@@ -21,18 +21,28 @@ import { QueryErrorComponent } from './query-error/query-error.component';
...
   private readonly changeDetectorRef = inject(ChangeDetectorRef);
 
+  ngOnInit() {
+    this.subscriptions.push(
+      this.validationService.validationPassed$.subscribe((validationPassed: boolean) => {
+        this.validationPassed = validationPassed;
+        this.changeDetectorRef.detectChanges();
+      }),
+    );
+  }
```
Helias, line 44: this is not necessary, you could use the `AsyncPipe` in the HTML template, like

```html
[disabled]="!(validationService.validationPassed$ | async)"
```

Remember to add `AsyncPipe` in the `imports: []` of this component

libs/shared/common-services/src/validation.service.ts
```diff
@@ -0,0 +1,9 @@
+import { Injectable } from '@angular/core';
+import { BehaviorSubject } from 'rxjs';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class ValidationService {
+  public validationPassed$: BehaviorSubject<boolean> = new BehaviorSubject(true);
```
Helias, line 8: public is not necessary, moreover you could make it constant using the `readonly` attribute
```suggestion
  readonly validationPassed$: BehaviorSubject<boolean> = new BehaviorSubject(true);
```

libs/shared/directives/src/validate-input.directive.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  ngOnInit(): void {
+    const control = this.ngControl.control;
+
+    if (!control) {
+      return;
+    }
+
+    this.subscriptions.push(
+      control.statusChanges?.pipe(distinctUntilChanged()).subscribe(() => {
```
Helias, line 27: This should improve a bit the performance
```suggestion
      control.statusChanges?.pipe(distinctUntilChanged(), debounceTime(500)).subscribe(() => {
```

Exitare: I added it, but it causes a noticeable lag because of the 500ms delay

Helias: we could set 200ms

libs/shared/base-abstract-classes/src/service/editors/editor.service.ts
```diff
@@ -82,10 +82,15 @@ export abstract class EditorService<T extends TableRow> extends SubscriptionHand
   }
 
   protected initForm(): void {
+    const defaultValues = new this._entityClass();
+
+    // Initialize the FormGroup
```
Helias, line 87:
```suggestion
```

libs/shared/base-abstract-classes/src/service/editors/single-row-editor.service.ts
```diff
@@ -86,19 +86,27 @@ export abstract class SingleRowEditorService<T extends TableRow> extends EditorS
 
   protected updateFormAfterReload() {
```
Helias, line 87: Not related to the PR
```suggestion
  protected updateFormAfterReload(): void {
```

libs/shared/base-abstract-classes/src/service/editors/single-row-editor.service.ts
```diff
@@ -86,19 +86,27 @@ export abstract class SingleRowEditorService<T extends TableRow> extends EditorS
...
-      const control = this._form.controls[field];
-      /* istanbul ignore else */
-      if (control) {
-        control.setValue(this._originalValue[field]);
-      } else {
-        console.error(`Control '${field}' does not exist!`);
-        console.log(`----------- DEBUG CONTROL KEYS:`);
-        for (const k of Object.keys(this._form.controls)) {
-          console.log(k);
+      // Ensure `field` is of type `string`
```
Helias, line 91: I usually avoid to write comments when the code is readable and clear
```suggestion
```

libs/shared/base-abstract-classes/src/service/editors/single-row-editor.service.ts
```diff
@@ -86,19 +86,27 @@ export abstract class SingleRowEditorService<T extends TableRow> extends EditorS
...
+        if (control) {
+          const value = this._originalValue[field as keyof T]; // Ensure type safety here
+          control.setValue(value as T[typeof field]);
+        } else {
+          console.error(`Control '${field}' does not exist!`);
+          console.log(`----------- DEBUG CONTROL KEYS:`);
+          for (const k of Object.keys(this._form.controls)) {
+            console.log(k);
+          }
         }
```
Helias, line 104: not sure about this code inside the else block, it seems more a debug code

libs/shared/common-services/src/validation.service.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+describe('ValidationService', () => {
+  let service: ValidationService;
+
+  beforeEach(() => {
+    TestBed.configureTestingModule({
+      providers: [ValidationService],
+    });
+
+    service = TestBed.inject(ValidationService);
+  });
```
Helias, line 14: we should use the setup() function strategy for initializing the service, for test isolation
see how the other tests are written and make the setup function please

libs/shared/directives/src/validate-input.directive.spec.ts
```diff
@@ -0,0 +1,174 @@
...
+    TestBed.configureTestingModule({
+      declarations: [TestComponent],
+      imports: [ReactiveFormsModule, FormsModule, InputValidationDirective],
+      providers: [ValidationService],
+    }).compileComponents();
+
+    fixture = TestBed.createComponent(TestComponent);
+    debugElement = fixture.debugElement.query(By.directive(InputValidationDirective));
+    fixture.detectChanges();
+  });
```
Helias, line 35: use the setup strategy

libs/shared/directives/src/validate-input.directive.ts
```diff
@@ -0,0 +1,65 @@
...
+
+    if (control?.invalid) {
+      control.markAsTouched();
+    }
+
+    if (control?.touched && control?.invalid && control.errors) {
+      const parent = this.el.nativeElement.parentNode;
+      if (parent) {
+        this.errorDiv = this.renderer.createElement('div');
+        this.renderer.addClass(this.errorDiv, 'error-message');
```
Helias, line 55: I think that I wrote this, but I am not sure if we have a class `error-message`, if not, we could remove this line of code

Exitare: We dont. But the tests use it to get the div in order to verify the correct message.

Helias: I would use an id instead of a class for this

Helias (approved):

libs/features/creature/src/creature-template/creature-template.component.ts
```diff
@@ -59,12 +60,13 @@ import { RouterLink } from '@angular/router';
...
     SpellSelectorBtnComponent,
     CreatureSelectorBtnComponent,
-    Model3DViewerComponent,
     GenericOptionSelectorComponent,
     BooleanOptionSelectorComponent,
     IconSelectorComponent,
     RouterLink,
+    InputValidationDirective,
   ],
+  providers: [ValidationService],
```
Francesco-Borzi, line 69: why does it have to be singleton?

libs/shared/base-abstract-classes/src/service/editors/editor.service.ts
```diff
@@ -82,10 +82,14 @@ export abstract class EditorService<T extends TableRow> extends SubscriptionHand
...
   protected initForm(): void {
+    const defaultValues = new this._entityClass();
+
     this._form = new FormGroup<ModelForm<T>>({} as any);
 
+    // Loop through the fields and initialize controls with default values
     for (const field of this.fields) {
-      this._form.addControl(field, new FormControl());
+      const defaultValue = defaultValues[field];
+      this._form.addControl(field, new FormControl(defaultValue, [Validators.required]));
```
Francesco-Borzi, line 92:
```suggestion
      this._form.addControl(field, new FormControl(defaultValues[field], [Validators.required]));
```

libs/shared/base-abstract-classes/src/service/editors/editor.service.ts
```diff
@@ -82,10 +82,14 @@ export abstract class EditorService<T extends TableRow> extends SubscriptionHand
   }
 
   protected initForm(): void {
+    const defaultValues = new this._entityClass();
+
     this._form = new FormGroup<ModelForm<T>>({} as any);
 
+    // Loop through the fields and initialize controls with default values
```
Francesco-Borzi, line 89: I don't think we need this comment, IMHO the code seems already quite self-explanatory :)

```suggestion
```

libs/shared/base-abstract-classes/src/service/editors/single-row-editor.service.ts
Francesco-Borzi: I agree

libs/shared/base-abstract-classes/src/service/editors/single-row-editor.service.ts
Francesco-Borzi: @Exitare please do not mark comments as resolved without answering or addressing them

libs/shared/base-abstract-classes/src/service/editors/single-row-editor.service.ts
```diff
@@ -80,29 +80,31 @@ export abstract class SingleRowEditorService<T extends TableRow> extends EditorS
...
-      if (control) {
-        control.setValue(this._originalValue[field]);
-      } else {
-        console.error(`Control '${field}' does not exist!`);
-        console.log(`----------- DEBUG CONTROL KEYS:`);
-        for (const k of Object.keys(this._form.controls)) {
-          console.log(k);
+      // Ensure `field` is of type `string`
+      if (typeof field === 'string') {
+        const control = this._form.controls[field];
```
Francesco-Borzi, line 93: is this change needed for the original purpose of the PR? in general, I'd prefer to split changes in several PRs, for example:

- 1 PR to include new functionality (including its unit tests)
- a separate PR to improve/refactor existing code that is not related to the functionality being introduced

libs/shared/base-editor-components/src/query-output/query-output.component.html
```diff
@@ -24,16 +24,40 @@
...
+    >
       <i class="fa fa-copy fa-sm"></i> {{ 'COPY' | translate }}
     </button>
-    <button type="button" class="btn btn-primary btn-sm" (click)="execute()" id="execute-btn">
+    <button
+      type="button"
+      class="btn btn-primary btn-sm"
+      (click)="execute()"
+      id="execute-btn"
+      [disabled]="(validationService.validationPassed$ | async) === false"
```
Francesco-Borzi, line 41: I think we should avoid this repetition and wrap `validationService.validationPassed$` in a variable. Even better to use a Signal.

libs/shared/directives/project.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "keira-shared-directives",
```
Francesco-Borzi, line 2: I think it's better to create a lib for validation instead of creating one lib for "directives" that can contain several directives that are unrelated to them.

So in the `validation` lib you can have all generic tools that are being used for the validation, including the validation service.

Francesco-Borzi (changes requested):

libs/features/creature/src/creature-template/creature-template.component.ts
Exitare: Because validation should happen on a page by page basis and not on the app level ?

libs/shared/base-editor-components/src/query-output/query-output.component.html
Exitare: I don't know how this requested change should look like.

libs/shared/base-abstract-classes/src/service/editors/single-row-editor.service.ts
Exitare: I resolved it, because I removed it initially. But then one of the tests failed and I added it back, but didn't unresolved the issue here.

libs/features/creature/src/creature-template/creature-template.component.ts
Helias: It should not be necessary, we could keep one instance for the entire application, so provide it in "root" and don't put it here.

If you find a strong reason to provide it inside the component then we could put it back.

libs/shared/base-editor-components/src/query-output/query-output.component.html
Helias: Before it was a variable with a subscription running on ngOninit, this affects the component adding a dependency for the OnInit interface and moreover, it forces you to manage the subscribe/unsubscribe.
In the current situation you don't need ngOninit and to manage the subscribe/unsubscribe thanks to the async pipe.

Wrapping this into a variable can be done without ngOninit still and in the following way:
```typescript
protected readonly isValidationPassed$ =  this.validationService.validationPassed$.pipe(map((val) => val === false));
```

Keeping in the template
```html
[disabled]="isValidationPassed$ | async"
```

Or, you could use signals

Exitare: OK, for learning purposes. What's the undesirable part about having a dependency in the ngOnInit, besides the subscription management part. Is it a general best practice not to use the ngOnInit if possible?

Helias: Personally I avoid any kind of dependencies, so if we can avoid ngOnInit, ngOnChanges etc. in an elegant way I would not add them. It increases the complexity of the logic component

Exitare:
Yes, I plan to.  I will be able to spend more time on the project again in around 2 months. If thats too long, I am fine with closing the PR.

FrancescoBorzi:
Closing so the bots stop reviewing this. Feelf free to reopen when you're back

Exitare:
I am slowly starting to work on this again

PR #3319, opened 2025-05-27
feat: integrate Unused-Guid-Search

sogladev:
This integrates https://github.com/Kitzunu/Unused-Guid-Search as a tab

<details><summary>UI Screenshot</summary>

keira3_unused

</details>

I copied the minimal UI, and the feature supports the same tables
- UI https://github.com/Kitzunu/Unused-Guid-Search/blob/master/guid2.PNG
- tables https://github.com/Kitzunu/Unused-Guid-Search/blob/aef451e81e6c1333bd89535f41d4ed4b9931372f/UnusedGuidSearcher/SearchGuids.cs#L13

It's still missing tests and translations, but I would prefer some feedback before, as I'm not too familiar with Angular architecture

### DIFF
--- a/libs/features/unused-guid-search/src/unused-guid-search.component.html
+++ b/libs/features/unused-guid-search/src/unused-guid-search.component.html
@@ -1,0 +1,58 @@
+<div class="container-fluid">
+  <div class="content-block">
+    <form (ngSubmit)="onSearch()" #guidForm="ngForm" class="mb-3">
+      <div class="form-group row">
+        <label class="col-sm-2 col-form-label">Table</label>
+        <div class="col-sm-4">
+          <select class="form-control" [(ngModel)]="selectedDb" name="selectedDb" required>
+            @for (db of dbOptions; track db) {
+              <option [ngValue]="db" [label]="db.label"></option>
+            }
+          </select>
+        </div>
+      </div>
+      <div class="form-group row mt-2">
+        <label class="col-sm-2 col-form-label">Start Index</label>
+        <div class="col-sm-4">
+          <input
+            type="number"
+            class="form-control"
+            [(ngModel)]="startIndex"
+            name="startIndex"
+            min="1"
+            [max]="MAX_INT_UNSIGNED_VALUE"
+            required
+          />
+        </div>
+      </div>
+      <div class="form-group row mt-2">
+        <label class="col-sm-2 col-form-label">Amount</label>
+        <div class="col-sm-4">
+          <input type="number" class="form-control" [(ngModel)]="amount" name="amount" min="1" [max]="MAX_INT_UNSIGNED_VALUE" required />
+        </div>
+      </div>
+      <div class="form-group row mt-2">
+        <label class="col-sm-2 col-form-label">Consecutive</label>
+        <div class="col-sm-4 d-flex align-items-center">
+          <input type="checkbox" [(ngModel)]="consecutive" name="consecutive" />
+        </div>
+      </div>
+      <div class="form-group row mt-3">
+        <div class="col-sm-6">
+          <button class="btn btn-primary" type="submit" [disabled]="loading">Search</button>
+        </div>
+      </div>
+    </form>
+
+    @if (loading) {
+      <div class="alert alert-info">Searching...</div>
+    }
+    @if (error) {
+      <div class="alert alert-danger">{{ error }}</div>
+    }
+    @if (results.length > 0) {
+      <div class="alert alert-success">Found unused GUIDs:</div>
+      <textarea id="results" class="mb-1" [value]="results.join(', ')" rows="6" spellcheck="false" readonly></textarea>
+    }
+  </div>
+</div>
--- a/libs/features/unused-guid-search/src/unused-guid-search.component.ts
+++ b/libs/features/unused-guid-search/src/unused-guid-search.component.ts
@@ -1,0 +1,134 @@
+import { Component, ChangeDetectionStrategy, ChangeDetectorRef } from '@angular/core';
+import { FormsModule, ReactiveFormsModule } from '@angular/forms';
+import { TranslateModule } from '@ngx-translate/core';
+import { CommonModule } from '@angular/common';
+import { MysqlQueryService } from '@keira/shared/db-layer';
+import { firstValueFrom } from 'rxjs';
+import {
+  CREATURE_SPAWN_TABLE,
+  CREATURE_SPAWN_ID_2,
+  GAMEOBJECT_SPAWN_TABLE,
+  GAMEOBJECT_SPAWN_ID_2,
+  CREATURE_TEMPLATE_TABLE,
+  CREATURE_TEMPLATE_ID,
+  GAMEOBJECT_TEMPLATE_TABLE,
+  GAMEOBJECT_TEMPLATE_ID,
+  GOSSIP_MENU_TABLE,
+  GOSSIP_MENU_ID,
+  NPC_TEXT_TABLE,
+  NPC_TEXT_ID,
+  PAGE_TEXT_TABLE,
+  PAGE_TEXT_ID,
+  ACORE_STRING_TABLE,
+  ACORE_STRING_ENTRY,
+} from '@keira/shared/acore-world-model';
+
+@Component({
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  selector: 'keira-features-unused-guid-search',
+  templateUrl: './unused-guid-search.component.html',
+  styleUrl: './unused-guid-search.component.scss',
+  imports: [CommonModule, FormsModule, ReactiveFormsModule, TranslateModule],
+})
+export class UnusedGuidSearchComponent {
+  readonly MAX_INT_UNSIGNED_VALUE = 4294967295;
+
+  dbOptions = [
+    { table: CREATURE_SPAWN_TABLE, key: CREATURE_SPAWN_ID_2, label: `${CREATURE_SPAWN_TABLE} (${CREATURE_SPAWN_ID_2})` },
+    { table: GAMEOBJECT_SPAWN_TABLE, key: GAMEOBJECT_SPAWN_ID_2, label: `${GAMEOBJECT_SPAWN_TABLE} (${GAMEOBJECT_SPAWN_ID_2})` },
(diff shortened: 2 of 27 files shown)


libs/features/unused-guid-search/src/unused-guid-search.component.ts
```diff
@@ -0,0 +1,134 @@
...
+  loading = false;
+  error = '';
+  consecutive = false;
+  amount = 10;
+  startIndex = 1;
+
+  constructor(
+    private mysql: MysqlQueryService,
+    private cdr: ChangeDetectorRef,
+  ) {}
```
Helias, line 60:
```suggestion
  private readonly mysql = inject(MysqlQueryService);
  private readonly cdr = inject(ChangeDetectorRef);
```

Helias:
I made some minor refactoring.

We usually don't use async/await but we work with rxjs, I see that you are using `firstValueFrom` to avoid rxjs and use async,await. If for you it's better to work with this, you can keep it as it is.
Moreover, I see that you are using ngModel, we are using more the FormGroup and FormControls in Keira3 so I would use them, but if you want to keep everything simple as it is to speed up the development I don't mind, especially because the forms will be converted soon in "SignalForms" in Angular, so sooner or later we'll refactor them.

About other variables, I converted them into signals, and removed the cdr change detection trigger, it is not necessary if you use signals.
Signals are a specific Angular features wich improve the application performances, avoid memory leak, set memorization automatically and many other features, so if possible I would still use them.

For the rest, the job seems a good start point!

sogladev:
@Helias, Thanks for the feedback and the improvements. I've refactored as needed to match the rest of the project, and I'lll look into setting up tests sometime in the future.

libs/features/unused-guid-search/src/unused-guid-search.component.ts
```diff
@@ -0,0 +1,152 @@
...
+    { table: 'pool_template', key: 'entry', label: 'pool_template (entry)' },
+    { table: 'game_event', key: 'eventEntry', label: 'game_event (eventEntry)' },
+  ];
+
+  protected form: FormGroup = this.fb.group({
+    selectedDb: [this.dbOptions[0], Validators.required],
+    startIndex: [1, [Validators.required, Validators.min(1), Validators.max(this.MAX_INT_UNSIGNED_VALUE)]],
+    amount: [10, [Validators.required, Validators.min(1), Validators.max(this.MAX_INT_UNSIGNED_VALUE)]],
+    consecutive: [false, [Validators.required]],
+  });
```
Helias, line 56: you could use `protected readonly form: FormGroup<FormGroupINTERFACE>`, so let's define an interface with the form group content.
Moreover I would avoid to use the formBuilder, use `new FormGroup<FormGroupINTERFACE>({ .... })`

libs/features/unused-guid-search/src/unused-guid-search.component.ts
```diff
@@ -0,0 +1,152 @@
...
+
+  protected results: string[] = [];
+  protected loading = signal(false);
+  protected error = signal('');
```
Helias, line 60: I forgot to set the signals as readonly

libs/features/unused-guid-search/src/unused-guid-search.component.ts
```diff
@@ -0,0 +1,152 @@
...
+@Component({
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  selector: 'keira-features-unused-guid-search',
+  templateUrl: './unused-guid-search.component.html',
+  styleUrl: './unused-guid-search.component.scss',
+  imports: [FormsModule, ReactiveFormsModule, TranslateModule],
+})
+export class UnusedGuidSearchComponent {
+  private readonly mysql = inject(MysqlQueryService);
+  private readonly fb = inject(FormBuilder);
```
Helias, line 33: the FormBuilder is not necessary

Helias (approved):

libs/features/unused-guid-search/src/unused-guid-search.component.spec.ts
```diff
@@ -0,0 +1,129 @@
+import { ComponentFixture, TestBed, waitForAsync } from '@angular/core/testing';
+import { of, throwError } from 'rxjs';
+import { UnusedGuidSearchComponent } from './unused-guid-search.component';
+import { MysqlQueryService } from '@keira/shared/db-layer';
+
+describe('UnusedGuidSearchComponent', () => {
+  let fixture: ComponentFixture<UnusedGuidSearchComponent>;
+  let queryService: MysqlQueryService;
```
FrancescoBorzi, line 8: let's avoid shared variables across tests and instead initialize and return these from the `setup` function (like you already do with the `component` object)

tsconfig.base.json
```diff
@@ -55,6 +56,7 @@
       "@keira/shared/switch-language": ["libs/shared/switch-language/src/index.ts"],
       "@keira/shared/test-utils": ["libs/shared/test-utils/src/index.ts"],
       "@keira/shared/utils": ["libs/shared/utils/src/index.ts"],
+      "keira-features-unused-guid-search": ["libs/features/unused-guid-search/src/index.ts"],
       "texts": ["libs/features/texts/src/index.ts"]
```
FrancescoBorzi, line 60: we should make it consistent and use `@keira/features/...` like in the other places

libs/features/unused-guid-search/src/unused-guid-search.component.spec.ts
```diff
@@ -0,0 +1,129 @@
...
+    const { component } = setupTest([{ guid: 1 }, { guid: 2 }, { guid: 4 }, { guid: 8 }, { guid: 9 }, { guid: 10 }]);
+
+    component['form'].patchValue({
+      selectedDb: component['dbOptions'][0], // creature
+      startIndex: -1,
+      amount: 10,
+      consecutive: true,
+    });
+    component['onSearch']();
+    expect(component['error']()).not.toBe('');
```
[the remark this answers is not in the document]
sogladev: I updated the button, loading, errors to use the library. I skipped the form as I wasn't sure how to approach it

```
component['form'].patchValue({
      selectedDb: component['dbOptions'][0], // creature
      startIndex: -1,
      amount: 10,
      consecutive: true,
    });
```

libs/features/unused-guid-search/src/unused-guid-search.component.html
```diff
@@ -0,0 +1,52 @@
+<div class="container-fluid">
+  <div class="content-block">
+    <form [formGroup]="form" (ngSubmit)="onSearch()" class="mb-3">
+      <div class="form-group row">
+        <label class="col-sm-2 col-form-label">Table</label>
```
Helias, line 5: We could put all the strings in en.json and leave them open for potential translations

libs/features/unused-guid-search/src/unused-guid-search.component.spec.ts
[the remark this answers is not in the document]
FrancescoBorzi: one way would be getting the `form` object not via `component['form']` but by getting it from the template instead, so in that case the test would also be testing that the form is properly integrated with the template. You can check:

https://francescoborzi.github.io/ngx-page-object-model/docs/advanced/testing-form-controls

please let me know if this is clear!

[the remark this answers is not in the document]
sogladev: The article is clear on how to access the component similar to `component['form']`, and asserting that initial values are set correctly. But I am missing the equivalent of `patch_values()`. I tried `page.form.patchValue({
` but it fails :smile:  I feel it should be similar to the button being called through the page object. Should `page.SetInputValueById()` be used, and every input field be marked by a `data-testid` and be set individually? It's complicated because it's not just a simple string FormControl but a FormGroup with multiple types

code that fails all the tests ```Error: NG0303: Can't set value of the 'formControl' input on the 'UnusedGuidSearchComponent' component. Make sure that the 'formControl' property is annotated with @Input() or a mapped @Input('formControl') exists.```
https://github.com/sogladev/Keira3/compare/feat-unused-guids..feat-unused-guids-forms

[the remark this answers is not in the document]
FrancescoBorzi: @sogladev I have prepared something - https://github.com/FrancescoBorzi/ngx-page-object-model/pull/104

I will let you know once it's available in Keira3 as well.

libs/features/unused-guid-search/src/unused-guid-search.component.html
```diff
@@ -0,0 +1,62 @@
+<div class="container-fluid">
+  <div class="content-block">
+    <form data-testid="form" [formGroup]="form" (ngSubmit)="onSearch()" class="mb-3">
```
Francesco-Borzi, line 3:
```suggestion
    <form data-testid="unused-guid-search-form" [formGroup]="form" (ngSubmit)="onSearch()" class="mb-3">
```

libs/features/unused-guid-search/src/unused-guid-search.component.spec.ts
```diff
@@ -0,0 +1,149 @@
...
+  get loading(): HTMLDivElement | null {
+    const debugEl = this.getDebugElementByTestId<HTMLDivElement>('loading', false);
+    return debugEl ? debugEl.nativeElement : null;
+  }
+  get errors(): HTMLDivElement | null {
+    const debugEl = this.getDebugElementByTestId<HTMLDivElement>('errors', false);
+    return debugEl ? debugEl.nativeElement : null;
+  }
+  get form(): FormGroup {
+    const debugEl = this.getDebugElementByTestId('form');
```
Francesco-Borzi, line 23:
```suggestion
    const debugEl = this.getDebugElementByTestId('unused-guid-search-form');
```

libs/features/unused-guid-search/src/unused-guid-search.service.ts
```diff
@@ -0,0 +1,89 @@
...
+export type DbOptions = {
+  table: string;
+  key: string;
+  label: string;
+};
+
+@Injectable({
+  providedIn: 'root',
+})
+export class UnusedGuidService {
```
FrancescoBorzi, line 16: Missing unit tests for this service

FrancescoBorzi: @sogladev ok, however it would still be good to have a placeholder test even for just creating the `service` and asserting that it is defined. I know it might sound useless but sometimes this kind of tests can find issues, for example when one does big refactoring or other major technical changes.

Helias (approved):

PR #3396, opened 2025-07-03
fix(features/gameobject): display correct type name

sogladev:
### Changes
GameObject type Select displays the correct enum name

culprit: 
- https://github.com/azerothcore/Keira3/blob/1c7537f14029474f6bdc1ee28c13f59dcc4ea124/libs/features/gameobject/src/select-gameobject/select-gameobject.component.ts#L40

### How to test
1. gameobject tab, select
2. search for door
3. See that type name is off by 1

keira_0_should_be_door


Helias:
thank you!

PR #3459, opened 2025-08-05
fix(item): remove StatsCount field

Tereneckla:
The column got dropped and the server handles counting now
https://github.com/azerothcore/azerothcore-wotlk/commit/9ad7cef3c474faacd2999b8575cbfc79a4d58852


Tereneckla:
the test works locally on my side

FrancescoBorzi:
hi @Tereneckla thank you very much for this PR, the code looks good but the build fails because of an integration test.

As you can see from the CI output:

image

You can reproduce it by running `nx run keira-features-item:test --watch=false`

or just: `nx test keira-features-item`

Let me know if you need any assistance with fixing it!

Tereneckla:
I did, but the test is passing locally and here on github the expected outcome is the same as the result (at least the part that is shown)

Helias:
I think that I have fixed it

Tereneckla:
Number blindness strikes again

FrancescoBorzi (approved):

PR #3498, opened 2025-08-24
feat: 3d model viewer for any items via character view

Helias:


libs/shared/model-3d-viewer/src/model-3d-viewer.component.spec.ts
```diff
@@ -52,17 +67,33 @@ describe('Model3DViewerComponent', () => {
...
-    component.ngOnChanges({ displayId: { currentValue: 1, previousValue: 0 } as any });
+      expect(resetModel3dElementSpy).toHaveBeenCalledTimes(0);
+      expect(show3DmodelSpy).toHaveBeenCalledTimes(0);
+    });
+
+    it('should call show3Dmodel if jQuery is present and displayId changed', (done) => {
+      const { component } = setup();
+      (component as any).windowRef.jQuery = () => {};
+      (component as any).displayId = () => 2;
+      (component as any).viewerType = () => VIEWER_TYPE.NPC;
```
FrancescoBorzi, line 89: same as aove

package.json
```diff
@@ -125,6 +125,7 @@
     "karma-time-stats-reporter": "0.1.0",
     "lint-staged": "16.1.5",
     "lodash.isequal": "4.5.0",
+    "ng-mocks": "14.13.5",
```
FrancescoBorzi, line 128: not sure if it's a good idea to add this and, if so, let's do it in another PR anyway

Issue #3500, opened 2025-08-26
[Critical] - Search by entity of non-existing entries will cause high usage of the CPU


TheSCREWEDSoftware:
SmartAI -> Search by entity -> creature (or any option) -> 123456 -> Edit SmartAI

Check the "Toggle Developer Tools" (right click any where within keira) and see your CPU usage in your Task Mange or Resource Monitor

<img width="569" height="476" alt="Image" src="https://github.com/user-attachments/assets/28f7eaec-a328-4b15-aabf-17a4f0a0d5b9" />

Threads | CPU | Usage %(the last 3)
<img width="620" height="104" alt="Image" src="https://github.com/user-attachments/assets/d68d1e8f-696e-49b3-8a7e-544d8026233a" />

Creature
<img width="657" height="90" alt="Image" src="https://github.com/user-attachments/assets/40e62e33-c3f2-4b37-a272-93344ff43be5" />

<img width="413" height="61" alt="Image" src="https://github.com/user-attachments/assets/715ce0a7-71b6-4f3c-9eec-b8746982e2b7" />

Gameobject
<img width="656" height="182" alt="Image" src="https://github.com/user-attachments/assets/b4f3814a-db03-441a-9552-223c82ff900a" />

<img width="434" height="72" alt="Image" src="https://github.com/user-attachments/assets/769b1888-0611-4317-bd24-bbf1d0865115" />

Area Trigger doesnt seem to loop errors but doesnt know the source type either
<img width="570" height="87" alt="Image" src="https://github.com/user-attachments/assets/a01dab26-b16f-42a3-9dcc-db187910aa8f" />

<img width="383" height="64" alt="Image" src="https://github.com/user-attachments/assets/1ab7fdc7-5a2b-4f11-b541-c437de2aad66" />

Timed Actionlist
<img width="645" height="151" alt="Image" src="https://github.com/user-attachments/assets/441375b3-3eaf-4650-9d80-239e5c48d5bb" />

<img width="501" height="72" alt="Image" src="https://github.com/user-attachments/assets/132c726f-e244-445b-a76e-6b1bd0043f46" />

TheSCREWEDSoftware:
CC @FrancescoBorzi and @Helias

PR #3533, opened 2025-09-15
Only run the query when data changes

TheSCREWEDSoftware:
Closes: https://github.com/azerothcore/Keira3/issues/3500

Prior to this you would run TOO MANY queries (at the end of this PR) in milliseconds

- For `libs\shared\sai-editor\src\sai-editor.component.html` 

Now uses `selectedName$` instead of `getHandler().getName()`

- `libs\shared\sai-editor\src\sai-editor.component.ts`

Now "caches" the value being looked/searched and only updates / run the query if any changes happen instead of checking multiple times a milisecond.
`SELECT ct.name FROM creature_template AS ct INNER JOIN creature AS c ON c.id1 = ct.entry WHERE c.guid = $ID`

https://gist.github.com/TheSCREWEDSoftware/52981434a8615ae770e638125071df97

### DIFF
--- a/libs/shared/sai-editor/src/sai-editor.component.ts
+++ b/libs/shared/sai-editor/src/sai-editor.component.ts
@@ -67,6 +67,8 @@ import { MultiRowEditorComponent } from '@keira/shared/base-abstract-classes';
 import { FlagsSelectorBtnComponent } from '@keira/shared/selectors';
 import { EditorButtonsComponent, QueryOutputComponent } from '@keira/shared/base-editor-components';
 import { AsyncPipe } from '@angular/common';
+import { Observable } from 'rxjs';
+import { shareReplay } from 'rxjs/operators';
 
 @Component({
   changeDetection: ChangeDetectionStrategy.OnPush,
@@ -90,6 +92,13 @@ import { AsyncPipe } from '@angular/common';
 export class SaiEditorComponent extends MultiRowEditorComponent<SmartScripts> implements OnInit {
   public override readonly editorService = inject(SaiEditorService);
   protected override readonly handlerService = inject(SaiHandlerService);
+
+  selectedName$!: Observable<string>;
+
+  override ngOnInit(): void {
+    super.ngOnInit();
+    this.selectedName$ = this.handlerService.getName().pipe(shareReplay(1));
+  }
 
   readonly EVENT_PHASE_MASK = EVENT_PHASE_MASK;
   readonly SMART_EVENT_FLAGS = SMART_EVENT_FLAGS;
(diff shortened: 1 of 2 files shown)


TheSCREWEDSoftware:
Only does this once 
image

Until updated

TheSCREWEDSoftware:
CC @FrancescoBorzi

libs/shared/sai-editor/src/sai-editor.component.ts
```diff
@@ -91,6 +93,13 @@ export class SaiEditorComponent extends MultiRowEditorComponent<SmartScripts> im
   public override readonly editorService = inject(SaiEditorService);
   protected override readonly handlerService = inject(SaiHandlerService);
 
+  selectedName$!: Observable<string>;
+
+  override ngOnInit(): void {
+    super.ngOnInit();
+    this.selectedName$ = this.handlerService.getName().pipe(shareReplay(1));
```
FrancescoBorzi, line 100: I like the idea of using `shareReplay(1)`, however shall we try using it directly inside the sai-handler service?

something like:

```
    return this.queryService.query<{ name: string }>(query).pipe(
      map((data) => {
        if (data.length > 0) {
          return `${data[0].name}`;
        } else {
          console.error(`Unable to find name for source_type = ${sai.source_type}, entryorguid = ${sai.entryorguid}`);
          return '';
        }
      }),
      shareReplay(1), // <--- ADD THIS HERE
    );
```

then there is no need for changes in the component or in other consumers of that service

TheSCREWEDSoftware:
Un-related to this, on startup

```
[11488:0915/133059.679:ERROR:CONSOLE:2] "Uncaught (in promise) Error: Error invoking remote method 'BROWSER_GET_LAST_WEB_PREFERENCES': WebContents does not exist", source: node:electron/js2c/sandbox_bundle (2)
[11488:0915/133100.484:ERROR:CONSOLE:1] "Request Autofill.enable failed. {"code":-32601,"message":"'Autofill.enable' wasn't found"}", source: devtools://devtools/bundled/core/protocol_client/protocol_client.js (1)
[11488:0915/133100.484:ERROR:CONSOLE:1] "Request Autofill.setAddresses failed. {"code":-32601,"message":"'Autofill.setAddresses' wasn't found"}", source: devtools://devtools/bundled/core/protocol_client/protocol_client.js (1)
```

Doing the way you asked, not i've to figure out how to cache the observable, otherwise just has the same behaviour as before

image

Francesco-Borzi (approved):

FrancescoBorzi:
very good, thanks @TheSCREWEDSoftware

Helias:
very nice!

Issue #3556, opened 2025-09-25
Connection to MySQL Fails When Secure Transport is Enabled


preeminence87:
I have MySQL running in a docker container on version 9.2.0-1.el9. In the [mysqld] stanza in my configuration, I have require_secure_transport=ON along with internally signed and issued certificate keypair:

[mysqld]
require_secure_transport=ON
ssl_ca=/var/lib/mysql/certs/myAuthority.crt
ssl_cert=/var/lib/mysql/certs/mysqlCert.crt
ssl_key=/var/lib/mysql/certs/mysqlCert.key

This configuration ensures the MySQL connecting clients are using secure transport, and everything is working for my worldservers.

However, here's the error returned from Keira 3.10.2 when connecting to MySQL with secure transport enabled:

```
Error:	Error: Connections using insecure transport are prohibited while --require_secure_transport=ON. at Packet.asError (C:\Users\<redacted>\AppData\Local\Temp\32jFdkEdq8gWZIvZK9H528iO3oy\resources\app.asar\node_modules\mysql2\lib\packets\packet.js:740:17) at ClientHandshake.execute (C:\Users\<redacted>\AppData\Local\Temp\32jFdkEdq8gWZIvZK9H528iO3oy\resources\app.asar\node_modules\mysql2\lib\commands\command.js:29:26) at Connection.handlePacket (C:\Users\<redacted>\AppData\Local\Temp\32jFdkEdq8gWZIvZK9H528iO3oy\resources\app.asar\node_modules\mysql2\lib\base\connection.js:475:34) at PacketParser.onPacket (C:\Users\<redacted>\AppData\Local\Temp\32jFdkEdq8gWZIvZK9H528iO3oy\resources\app.asar\node_modules\mysql2\lib\base\connection.js:93:12) at PacketParser.executeStart (C:\Users\<redacted>\AppData\Local\Temp\32jFdkEdq8gWZIvZK9H528iO3oy\resources\app.asar\node_modules\mysql2\lib\packet_parser.js:75:16) at Socket.<anonymous> (C:\Users\<redacted>\AppData\Local\Temp\32jFdkEdq8gWZIvZK9H528iO3oy\resources\app.asar\node_modules\mysql2\lib\base\connection.js:100:25) at Socket.emit (node:events:519:28) at addChunk (node:internal/streams/readable:561:12) at readableAddChunkPushByteMode (node:internal/streams/readable:512:3) at Readable.push (node:internal/streams/readable:392:5)
[...]

PR #3558, opened 2025-09-25
feat: SSL/TLS implementation for require_secure_transport

TheSCREWEDSoftware:
This implementation allows people that have require_secure_transport enabled.

Closes: https://github.com/azerothcore/Keira3/issues/3556

https://github.com/azerothcore/Keira3/issues/3556#issuecomment-3335818479

Being tested by the issue author also.

I tried following the current config sturcutre that's being used, I'm unsure about the description for the SSL/TLS, im just thinking of having it without a description, open to feedback.

This can be tested if you want on your side by using your super user for mysql and running the following:
```sql
SET GLOBAL require_secure_transport = ON;
```

And having TLS option off.
Image

And having TLS option on

Image


TheSCREWEDSoftware:
CC @FrancescoBorzi and @Helias

TheSCREWEDSoftware:
This I may need help fixing, as I'm unsure now from the error, it does build fine in windows and works

libs/main/connection-window/src/connection-window.component.ts
```diff
@@ -75,13 +82,25 @@ export class ConnectionWindowComponent extends SubscriptionHandler implements On
...
     this.subscriptions.push(
-      this.mysqlService.connect(this.form.getRawValue()).subscribe({
+      this.mysqlService.connect(connectionConfig).subscribe({
         next: () => {
           const newConfig = this.form.getRawValue();
           if (!this.savePassword) {
             newConfig.password = '';
           }
+          // Save SSL preference with the config
+          (newConfig as any).sslEnabled = this.sslEnabled;
```
Helias, line 102: same as below

Helias (approved):

TheSCREWEDSoftware:
// SET GLOBAL require_secure_transport = ON;

Unchecked = Error, intended
image

Checked = No error, intended
image

And thanks helias for adjusting it to the Keira standards

libs/main/connection-window/src/connection-window.component.ts
```diff
@@ -52,7 +54,11 @@ export class ConnectionWindowComponent extends SubscriptionHandler implements On
 
     if (this.configs?.length > 0) {
       // get last saved config
-      this.form.setValue(this.configs[this.configs.length - 1]);
+      const lastConfig = this.configs[this.configs.length - 1];
+      this.form.setValue(lastConfig);
+
+      // Restore SSL preference if it was saved with the config
+      this.sslEnabled = !!lastConfig.sslEnabled;
```
FrancescoBorzi, line 61: it's not critical but why not just include `sslEnabled` inside the form instead of keeping it outside of it?

TheSCREWEDSoftware:
Can test this during the afternoon

TheSCREWEDSoftware:
All working

PR #3573, opened 2025-10-04
feat(option-icon): improve icon management, add several icons

Helias:


libs/shared/test-utils/src/lib/keira-page-object.ts
```diff
@@ -183,4 +183,24 @@ export abstract class PageObject<ComponentType> extends PageObjectModel<Componen
...
+    const ngxSelectEl = this.fixture.debugElement.nativeElement.querySelector(`#${id} ngx-select`);
+    const toggle = ngxSelectEl.querySelector('.ngx-select__toggle');
+    toggle.click();
+
+    this.detectChanges();
+
+    // select the option
+    this.getAllDebugElementsByCss(`#${id} ngx-select .ngx-select.open .ngx-select__item`)[index].nativeElement.click();
+
+    this.detectChanges();
```
FrancescoBorzi: it is not ideal to call `detectChanges`, especially twice (see [this](https://javascript.plainenglish.io/angular-zoneless-unit-testing-e048e1d9220d)), can you try replacing it with just `await this.whenStable();` and making the method async?

FrancescoBorzi (approved):

libs/shared/selectors/src/selectors/generic-option-selector/generic-option-selector.component.spec.ts
```diff
@@ -26,9 +26,10 @@ describe('GenericOptionSelectorComponent', () => {
     }).compileComponents();
   });
 
-  const setup = () => {
+  const setup = (optionList = EXPANSION) => {
```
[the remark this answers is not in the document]
Helias: I will do it in another occasion

libs/shared/test-utils/src/lib/keira-page-object.ts
Helias: it worked, removed both detectChanges :heavy_check_mark:

PR #3647, opened 2025-12-14
docs(spell): update `spell_dbc` documentation

sogladev:
> The spell_dbc table contains mostly overrides of the spells located in the Client DBC files.

This is no longer correct. `spell_dbc` no longer contains any client DBC data. This aims to correct the 'HEADS UP: this is how the spell_dbc table works' section on this page.


Helias:
Thank you!

Issue #3644, opened 2025-12-03
npc_trainer align AC


Helias:
https://github.com/azerothcore/azerothcore-wotlk/pull/23040

PR #3646, opened 2025-12-14
feat(create): add 'copy from existing' (clone) option to CreateComponent

sogladev:
### Changed proposed:

Add support for cloning an existing entity from the 'Create' card for multiple types.

I modified the tests for `create.component.spec.ts`. I'm not sure what other tests are needed for this.

FYI not an NX/Angular expert and I used AI tools to help draft this. Hopefully, nothing too bad in here 

### Details:

Adds `allowCopy` field to the CreateComponent `<keira-create>`, which is opt-in.

Handling is done by routing to a `/clone` route with a provided `ID` to copy from. This is similar to `/select` and `/create` that already exist. The `/copy` route displays a page with `copy` queries for the main table and related tables. After executing, the user is routed to the edit page.

After doing this for creatures, it was easy enough to copy and paste the handling for other types. 

For gobjects and creatures, only the template-like tables are copied. SmartAI (TODO?), spawns or guid-specific stuff is not copied.

### Issues

- Closes https://github.com/azerothcore/Keira3/issues/25
- Closes https://github.com/azerothcore/Keira3/issues/700
- Closes https://github.com/azerothcore/Keira3/issues/964

### UI 

<details> <summary> UI Screenshots </summary>

keira_38_1

keira_3

keira_4

</details>

### [[copilot summary below]]

This pull request adds support for "copy" routes and features across multiple entity types in the Keira app, allowing users to duplicate existing entries. It also introduces new internationalized strings related to the copy functionality in all supported languages.

Key changes include:

**Routing and Feature Enhancements:**

[...]

PR #3653, opened 2025-12-14
feat(trainer): add trainer support

sogladev:
Requires:
- https://github.com/azerothcore/azerothcore-wotlk/pull/23040

- Closes https://github.com/azerothcore/Keira3/issues/3644

This is missing tests, and the PR is incompatible with the below (the other needs to be updated if either is merged)
- https://github.com/azerothcore/Keira3/pull/3646


Helias:
I have added spells icons and the spellName column to have a better visualization
image

I am working on it

PR #3672, opened 2026-01-19
Update new SAI Actions

Gultask:
https://github.com/azerothcore/azerothcore-wotlk/pull/24442

Also changes a bug in an old action of mine + add comment handling to AnimTier and new action


Helias:
Thanks!

PR #3370, opened 2025-06-23
Refactor QuestPreviewComponent

FrancescoBorzi:
Generated with AI

Issue #3356, opened 2025-06-22
[Enchantement] - SQL Editor Improvements


TheSCREWEDSoftware:
Refering to: https://github.com/azerothcore/Keira3/releases/tag/v3.9.0

It's much better than before (having numbered lines, row background and colour syntax).

Currently missing is the suggestion of the table or column value (and different colour)

![Image](https://github.com/user-attachments/assets/2917182b-f5e4-4b13-a357-2306c3f15cca)
![Image](https://github.com/user-attachments/assets/68bcdfc7-a4a5-4fc0-8b30-580baf65c621)

![Image](https://github.com/user-attachments/assets/a656a2f0-2911-4f68-a8f4-72abb655eee8)
![Image](https://github.com/user-attachments/assets/960692fb-63c8-4742-a89f-d685cc934ce2)

the "USE" command and similiar commands (that related to interact with other databases than the one you're logged in / accessing) should be removed/forbidden/not readable, in the case of keira is doesnt come well format (nor shouldn't, as this is made for acore_world and not any of the other two.

![Image](https://github.com/user-attachments/assets/ab30bdf4-4fd9-42c5-9459-8b113fe47eae)
![Image](https://github.com/user-attachments/assets/62aed80a-4574-4123-a6e1-7b883300aaa2)

Vuelos:
Maybe solved on #3757

TheSCREWEDSoftware:
Just have to add "Closes issueLink" in the PR

Helias:
covered by https://github.com/azerothcore/Keira3/pull/3757

PR #3751, opened 2026-05-11
Quest preview objectives, sql editor changes

Vuelos:
Completes PR #3370
Tackles #3356: Adds Schema to autocomplete on the SQL Editor
Tackles #3174 and #3132: Adds Button to show/hide trace.

Tested on local instance of AzerothCore.

Test captures:

Parse Text with line breaks:
imagen
Creature reference:
imagen
Item reference:
imagen
Schema (TABLES and COLUMNS) on autocomplete
imagen
New Error message:
imagen
imagen


Vuelos:
I want to do a shared Map component (to show creatures spawns, location, etc.).
What do you think is the best approach to getting the maps images?. Is there an external source for them like the 3D models?

libs/features/quest/src/quest-preview/quest-objectives.component.ts
```diff
@@ -0,0 +1,35 @@
...
+export interface FactionRequirement {
+  name$: Observable<string>;
+  value: string;
+}
+
+@Component({
+  selector: 'keira-quest-objectives',
+  templateUrl: './quest-objectives.component.html',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  standalone: true,
```
Helias, line 26: this is not necessary anymore because by default the property standalone is true

Helias:
Hola! :wave: 

first, I strongly appreciate this PR, reviewing it quickly seems ok. If you can, would be great to split this it into 2 PRs, one with the quest preview improvements and the other one with the sql editor auto complete feature, so it's easier to review, test and merge, if you can't it's fine, I can test this and then merge.

About the **Map** question, I was wondering the same and some months ago I was investigating how AoWoW does it https://github.com/azerothcore/aowow  
I investigate it long time ago so now I do not remember exactly but, from my understanding, it should use WorldMapArea.dbc and it converts using some specific formula the gps coords that we have in the ac database into local map coords, and it should do it automatically during the "setup" process.

I would strongly recommend you to check azerothcore/aowow and extract the formula or give the two projects azerothcore/aowow and azerothcore/azerothcore-wotlk to an AI for extracting it.

let me know if you have further questions or if you want to talk about it, feel free to reach me via the discord server azerothcore / #keira3 channel or in DM

Helias:
I checked aowow quickly and you can find the implementation to transform the world coords into zone pos in `game.php` in the function `worldPosToZonePos()` which is used in spawns.ss.php file in the `transformPoint()` function

Vuelos:
Replaced by #3752 only including quest objective preview

PR #3758, opened 2026-05-15
feat(Claude): add angular skills and update .gitignore to support plans

FrancescoBorzi:


Vuelos:
Hi!
I'm using AI skills too, but instead of adding them directly on the projects I try to use git to link them from other repos.
I use a python script but you can also use git submodules or sparse checkouts.

Source repos you can use:
https://github.com/wondelai/skills
https://github.com/analogjs/angular-skills

I use the azerothcore wiki .md files as a skill too:
Powershell to clone using sparse checkout getting /docs/ only:

````powershell
git clone `
  --no-checkout `
  --depth=1 `
  --filter=tree:0 `
  https://github.com/azerothcore/wiki.git wiki

cd wiki

git sparse-checkout init --no-cone

@"
/docs/
!/docs/es/
"@ | Set-Content .git/info/sparse-checkout

git checkout
git pull
```
`

FrancescoBorzi:
@Vuelos interesting alternative approach, but how do you make sure all developers load all skills?

Vuelos:
> @Vuelos interesting alternative approach, but how do you make sure all developers load all skills?

We don't, we still consideer agent programming and AI skills usage optional.

FrancescoBorzi:
while it is still optional, I'd prefer that IF users use Agents, then they are fed with the right skills to operate on Keira3 :)

PR #3864, opened 2026-08-11
feat: quest chain view, condition markers and multi-row diff fixes

Gultask:
## What this does

Adds a **Quest Chain** visualiser, surfaces **condition links** across the Quest and SmartAI editors, and fixes the correctness bugs found while building them.

Happy to split this into separate PRs if maintainers prefer — the commit body groups the changes and they are largely independent.

### Quest Chain (new)

`libs/features/quest/src/quest-chain/` — a new sub-tab under Quest that renders the chain around the selected quest as an inline SVG graph.

- Three edge kinds: `PrevQuestID`, "enabled by" (negative `PrevQuestID`) and `BreadcrumbForQuestId`.
- `ExclusiveGroup` drawn as a band — positive = "one of these", negative = "all of these".
- Quests that are referenced but missing from `quest_template` are drawn as explicit **missing** nodes, so a chain that dead-ends shows you where it breaks. This was the motivating use case.
- Clicking a node selects that quest; a badge opens its availability conditions.
- BFS is capped (80 nodes / 25 rounds) with a truncation notice, and is cycle-safe. Layered layout, no new dependencies.

### Condition markers

- The SmartAI event table gains a `cond` column (between `flags` and `comment`) showing how many conditions are attached to each event, linking into the Conditions editor prefiltered.
- A new **Add new condition** button opens the Conditions editor for the selected event with its key already filled in.
- Quest chain nodes show the same marker for `CONDITION_SOURCE_TYPE_QUEST_AVAILABLE`.
- `SelectConditionsComponent` accepts query params to prefill the search and either run it or open the new-condition form. Additive — it no-ops without them.

`conditions.SourceGroup` is the `smart_scripts` id shifted by one (`GetConditionsForSmartEvent` looks up `eventId + 1`); a named constant and a test now record that.

## Bug fixes

**`conditions.SourceId` was dropped when creating a condition.** `ConditionMgr` keys smart-event conditions on `(SourceEntry, SourceId)`, so `SourceId` must equal `smart_scripts.source_type`. It was listed in `CONDITIONS_ID_FIELDS` but not in `CONDITIONS_SEARCH_FIELDS`, so the search form had no control for it, `getPartial` set it to `undefined`, and the new condition defaulted to `SourceId = 0` — silently attaching a gameobject script's condition to the creature script with the same `entryorguid`. It is now a search field and is carried on both condition links.

**The multi-row diff query could destroy data.** `updateDiffQuery()` passed only the secondary key to `getDiffDeleteInsertTwoKeysQuery()`, never `_entityExtraIdField`, and row matching returned the first row with that key. For a `creature_text` GroupID holding several ID variants — the standard "random variant" pattern — editing one variant produced:

```sql
DELETE FROM `creature_text` WHERE (`CreatureID` = 1234) AND (`GroupID` IN (0, 0));
INSERT INTO `creature_text` (...) VALUES
(1234, 0, 0, 'variant0', ...),
(1234, 0, 0, 'variant0', ...);
```

That deletes all three variants, re-inserts the first one twice, loses the edit, and violates the primary key. The `fullQuery` for the same state was correct, so the bug was confined to the diff path.

Row matching now uses the full key, involved rows are deduped, and a touched group is rewritten whole. The whole-group rewrite is gated on an extra id field being declared — that declaration is precisely the statement that the secondary key names a *group* rather than a row — so behaviour is byte-for-byte unchanged for every other editor.

**`quest_template_addon` was missing `BreadcrumbForQuestId`** (added upstream by `2026_01_29_00.sql`). Now editable, with a tooltip documenting the core's validation rules.

## Editor improvements

`creature_text`:
- New rows default to `Type` 12 (Say — `0` is not a valid chat type and is not in the dropdown) and `Probability` 100 (the core uses it as a weighted-random weight, so `0` makes the text unreachable).
- **Duplicate** now adds a variant to the *same* GroupID and advances the ID, instead of starting a new group. **Add new row** still starts a new group.

SmartAI:
- Added `SMART_ACTION_SUMMON_GAMEOBJECT_GROUP` (241), with its generated comment so the action list and comment map stay in step with the core.

## Testing

All touched libs pass and hold their coverage thresholds: sai-editor 308, db-layer 149, conditions 27, base-abstract-classes 77, creature 231, quest 219, item 203, selectors 101, gameobject 76, other-loots 72, smart-scripts 11. Lint clean, `npm run build` green.

New regression tests cover the multi-ID group diff (edit / delete one / delete whole group / add / combined across groups), the `SourceId` key including the `source_type = 0` case, and the creature_text add/duplicate id behaviour.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added an interactive Quest Chain view with relationships, statuses, conditions, legends, and navigation.
  * Added Quest Chain navigation and breadcrumb quest support.
  * Added Source ID condition filtering and deep-linkable condition searches.
  * Added SmartAI condition counts, creation, and navigation from event rows.
  * Added the “Summon GameObject Group” SmartAI action.

* **Improvements**
  * Creature text duplication now assigns group-specific IDs.
  * New creature text entries use valid default type and probability values.
  * Added English, Spanish, Russian, and Chinese translations.


Gultask:
Note, this is a vibecode. As in, I ask for a feature, and do not know the underlying changes necessary to make it work. I see it working, I vibed it as good.

I normally don't like vibecoding myself. But I also don't know TS, and my contributions to Keira have been very small so far, mostly in SAI comments and new Actions, which is just extra content to pre-existing foundations.

However, I am a frequent user of the software, and these changes do help me.

libs/features/quest/src/quest-chain/quest-chain.component.ts
```diff
@@ -0,0 +1,82 @@
...
+import { NODE_HEIGHT, NODE_WIDTH } from './quest-chain.layout';
+import { EMPTY_QUEST_CHAIN_GRAPH, QuestChainGraph, QuestChainNode } from './quest-chain.model';
+import { QuestChainService } from './quest-chain.service';
+
+@Component({
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  selector: 'keira-quest-chain',
+  templateUrl: './quest-chain.component.html',
+  styleUrls: ['./quest-chain.component.scss'],
+  imports: [TopBarComponent, TranslateModule],
```
Helias: do not use modules like TranslateModule, import only what you really need as standalone components/directives/pipes

libs/features/quest/src/quest-chain/quest-chain.component.ts
```diff
@@ -0,0 +1,82 @@
...
+})
+export class QuestChainComponent implements OnInit {
+  private readonly questChainService = inject(QuestChainService);
+  private readonly changeDetectorRef = inject(ChangeDetectorRef);
+  private readonly router = inject(Router);
+  readonly handlerService = inject(QuestHandlerService);
```
Helias: this should be protected instead of public

libs/features/quest/src/quest-chain/quest-chain.component.ts
```diff
@@ -0,0 +1,82 @@
...
+    try {
+      this.graph.set(await this.questChainService.buildGraph());
+    } catch (error) {
+      this.error.set(error instanceof Error ? error.message : String(error));
+      this.graph.set({ ...EMPTY_QUEST_CHAIN_GRAPH });
+    }
+
+    this.loading.set(false);
+    // Zoneless app: nothing else marks the view dirty after this await.
+    this.changeDetectorRef.markForCheck();
```
Helias: not sure if this is really necessary, I don't think so
changeDetectorRef should not be necessary in zoneless and OnPush component if we manage properly the reactivity

libs/shared/sai-editor/src/sai-editor.component.ts
```diff
@@ -185,4 +190,65 @@ export class SaiEditorComponent extends MultiRowEditorComponent<SmartScripts> im
...
+        // conditions.SourceGroup is the event id shifted by one; shift it back to match smart_scripts.id.
+        counts[Number(row.SourceGroup) - SMART_EVENT_CONDITION_GROUP_OFFSET] = Number(row.conditionCount);
+      }
+    } catch (error) {
+      // A supplementary marker must never take the editor down with it; just show no markers.
+      console.warn('Could not load SmartAI condition counts', error);
+    }
+
+    this.conditionCounts.set(counts);
+    this.changeDetectorRef.markForCheck();
```
Helias: same here

libs/shared/sai-editor/src/constants/sai-actions.ts
```diff
@@ -181,6 +181,7 @@ export enum SAI_ACTIONS {
   DISABLE_REWARD = 238,
   SET_ANIM_TIER = 239,
   SET_GOSSIP_MENU = 240,
+  SUMMON_GAMEOBJECT_GROUP = 241,
```
Helias, line 184: nit: we could do this in a separate PR

Helias:
> Note, this is a vibecode. As in, I ask for a feature, and do not know the underlying changes necessary to make it work. I see it working, I vibed it as good.
> 
> I normally don't like vibecoding myself. But I also don't know TS, and my contributions to Keira have been very small so far, mostly in SAI comments and new Actions, which is just extra content to pre-existing foundations.
> 
> However, I am a frequent user of the software, and these changes do help me.

that's fine and well accepted, I'll do some manual check when I can

Gultask:
Currently running a self-review, then doing a second pass overall

libs/features/quest/src/quest-chain/quest-chain.component.ts
FrancescoBorzi: we need to scaffold some project agent rules

Gultask:
Thanks @Helias — all four of your code comments are applied in `eb5f007d`:

- **`TranslateModule`** → `TranslateDirective` + `TranslatePipe`. I checked and that's already the pattern in the other standalone components here, so this was just me not following it.
- **`handlerService`** is now `protected`.
- **Both `markForCheck()` calls dropped**, along with the `ChangeDetectorRef` injection in `QuestChainComponent`. You're right — the state they were guarding is held in signals, and a signal write already schedules change detection when zoneless. Tests and a production AOT build still pass without them.
- **Action 241 in a separate PR** — happy to do that, and the same goes for the rest of this branch if you'd prefer it split. It's six fairly independent changes and the commits are grouped so splitting is mechanical. Say the word and I'll break it up rather than have you review it as one lump.

---

I also ran a self-review pass before you got to it, which turned up **six issues, all now fixed** in `2e93b954`. Two of them CodeRabbit found independently (the node budget and the band label), so those are covered too.

**Correctness:**

1. **Deep links leaked filters between screens.** `ConditionsSearchService` is app-lifetime, and `patchValue` only writes the keys it's given. So: click a `cond` badge in SmartAI (sets `SourceGroup` + `SourceId`), then click a condition badge on a quest — the quest search still carried the SmartAI `SourceGroup`, returned nothing, and "Create new" would then key the condition with those stale values. Same class of mis-keying the `SourceId` fix in this PR was for. Now resets before patching.
2. **Quest Chain "Reload" couldn't reload.** All three chain queries go through `queryToPromiseCached`. `EditorComponent` clears that cache on entry, but `QuestChainComponent` doesn't extend it, so Reload only re-ran the layout over cached rows. Now clears the cache.
3. **`truncated` wasn't set when the round limit stopped the walk** — only when the node budget did. A chain cut off mid-walk rendered with no warning, which is exactly what the "missing node" markers exist to avoid.
4. **`MAX_QUEST_CHAIN_NODES` didn't bound what was drawn** (CodeRabbit's nitpick). The check runs after a whole frontier is absorbed and the id list was never trimmed, so one wide round could lay out thousands of nodes through an O(nodes×edges) relaxation loop. Now applied to the drawn list, walk order preserved so the selected quest stays.

**Conventions:**

5. **Exclusive-group band captions were hardcoded English** rendered into the template (also CodeRabbit's). Moved into the template with `BAND_ANY`/`BAND_ALL` in all four locales.
6. **Two duplicated condition source-type constants** restating `CONDITION_SOURCE_TYPES` from `acore-world-model`. Deleted; consumers use the enum.

Each fix has a regression test that fails against the old code.

**On the remaining CodeRabbit comments**, for the record rather than to argue with a bot:

- The self-closing-tag one on `<i class="fa fa-filter fa-sm"></i>` I don't think is right — that's a native HTML element, and there are ~500 identical usages in the repo already.
- The accessibility points are fair. The `cond` badge being a clickable `<span>` should be a `<button>`; the SVG graph nodes need focus + keyboard handling, which is a bigger change. Both happy to do — tell me whether you want them here or as a follow-up.
- It also suggests rewriting the deep-link tests to navigate through the real router instead of overriding `ActivatedRoute`. I left them as-is since overriding the snapshot is the simpler and more common pattern, but I'll change it if you disagree.

Two things I want to be straight about: the second review pass got cancelled partway for cost, so `2e93b954` and `eb5f007d` haven't had a fresh-eyes check — just tests, lint and a production build, all green. And there's a separate bug I haven't been able to reproduce or fix, where creating a condition from a pristine search form crashes; it predates this branch and isn't touched by it.

Gultask:
> Two things I want to be straight about: the second review pass got cancelled partway for cost

I ran out of credits lol

Helias:
I am sorry that you finished your credits, I hope they will renew soon

btw, we are already tracking the quest chain here

image

your solution could give some more information, allowing you to navigate the single quest. In the future, we could do the same in that side and we could also add the additional information that you need there

Helias:
I have restyled it a bit

image

Gultask:
Yeah, I was going for the branching ones which can't be seen in a column.

image

Although I was also thinking if there were ways to expand the graph, so we can see, for example, what it unlocks.
Using conditions or, well, I don't really know 😭 

Like how they do in WoWHead sometimes:

image

I.e. the graph is mostly only useful to determine ExclusiveGroups, which are a bit of a nightmare to read, in my opinion.

Gultask:
Also, yeah, I was thinking on how to display it in the main sidebar, but the row-style and limited space make it a bit difficult to do so. It would be more of an "expanded" view.

Gultask:
It's a bit difficult because, well.

This is mostly just conjecture, I'm not sure what the *actual* ordering or limits are.
See this chain:

image

After it ends, it phases the player so they start seeing the questgivers for the next part of the Icecrown chain. But how would we show this in the graph? Do we read spell_dbc AND spell_area for the phasing data? Comb the conditions?
I don't really know if it's possible.

Maybe by going through conditions we can discover branches and side-chains, annotated by an (...) that indicates where it branches off.

Helias (approved):

Gultask:
I think I got overly excited and started adding more than I should :D

I still have some ideas, like adding a broadcast_text lookup into creature_text editor, but from now on it'll be separate PRs, my bad.

Gultask:
image

Got it. Conditions now checked as well. These first few quests would not show since they are only handled by conditions, not the template_addon

Gultask:
Check out the chain for Once More Unto The Breach, Hero (13104) now, lmao

Helias:
I'll merge this PR tomorrow or on thursday, thanks a lot for this feature ;-)

PR #3875, opened 2026-08-18
feat(creature): add broadcast_text browser to the creature_text editor

Gultask:
Search broadcast_text by MaleText or FemaleText from inside the creature_text tab, copy a result's id, language and text into the selected row, and pull in the adjacent ids when a search hits exactly one row.

I requested this feature because making blizzlike creature_text rows is very annoying.
Open creature_text
Open broadcast_text browser
Find text
Copy text, switch tab, paste text, switch tab, copy ID... you get it.

Also, one important feature is that sometimes the adjacent IDs are related, but you can't see it via the broadcast_text browser. This PR also shows adjacent rows if there's an exact match.

Example of usage:

https://github.com/user-attachments/assets/c055ea57-7422-4594-8f93-2a2961c80717

Running /self-review now

## Summary by CodeRabbit

* **New Features**
  * Added a broadcast-text browser for searching creature dialogue across male and female text.
  * Search results highlight matches and show nearby entries for context.
  * Copy a selected broadcast text’s ID, language, and wording into a creature-text row.
  * Added English, Spanish, Russian, and Chinese localization.
* **Bug Fixes**
  * Copying is disabled without a selected creature-text row.
  * Selected rows remain preserved when results refresh.
* **Release**
  * Releases now provide raw Linux, macOS, and Windows binaries with checksums.


Gultask:
# Self-review — creature-text-update → upstream/master

> **Outcome** 2 findings — 2 fixed — author stopped after round 1
> **⚠ Caveat** fixes not re-reviewed: round 2 declined by the author; head is now `fefd7c42`
> **Intent** Embed a broadcast_text browser in the creature_text tab so populating a row no longer needs tab-switching: search matching `MaleText OR FemaleText`, a button copying the picked row's language, ID and text into the selected creature_text row, and adjacent rows shown when the search hits exactly one.
> **Reviewed** `f2777496` (`creature-text-update` vs `upstream/master`, merge-base diff — 17 files, +625 −8)
> **By** Claude Code (Claude Agent SDK), Opus 5 — self-review v0.1, 2026-08-17
> **Project rules** `.agents/docs/self-review-rules.md` not present
> **Coverage** all 17 changed files reviewed

## Round 1 — `f2777496`, 2 findings

1. `libs/shared/db-layer/src/query/mysql-query.service.ts:590` — a `?` in the search text is read by squel as a
   bind placeholder and substituted with `undefined`, producing SQL MySQL rejects (`LIKE '%Who dares'undefined'%'`);
   `broadcast_text` is NPC dialogue, so question marks are routine → **fixed**: bound as a squel parameter,
   `escape` import dropped, regression test added
2. `libs/features/creature/src/creature-text/broadcast-text-browser/broadcast-text-browser.service.ts:8` — the
   root-scoped service outlives the component while the datatable's highlight does not, so after a tab or creature
   switch the copy button is armed by a selection the user cannot see, silently overwriting `Text`/`Language`/
   `BroadcastTextId` → **fixed**: `[selected]` bound to the service state so the two cannot diverge; re-creation test added

## Post-fix verification (not a review round)

`creature-text` 40 tests, db-layer query spec 110 tests, full creature lib 42 files / 249 tests at 100%
statements/branches/functions/lines; eslint (TS + templates) and `tsc` clean on both libs.

## CodeRabbit comments — assessed separately, no changes made

| # | Claim | Assessment |
|---|---|---|
| 1 | `en.json:327` "Id" → "ID" | Pre-existing line; the diff only adds a trailing comma. Out of scope |
| 2 | Search button submits the form and reloads the page | **Incorrect.** `FormGroupDirective.onSubmit` returns false for non-dialog forms, so Angular cancels the native submit; the identical pattern ships in `select-broadcast-text` and `select-npc-text` |
| 3 | `MaleText \|\| FemaleText` hides a distinct female wording | Real gap, but ruled out by the stated requirement: the two only differ when one side is blank, so one Text column was specified |
| 4 | Superseded searches can overwrite newer rows | Real in principle; identical to `SearchService.onSearch` repo-wide, slightly amplified by the two-stage query. Worst case is stale rows until the next search |
| 5 | `if (limit)` treats a cleared input as unlimited | **Not a defect.** Mirrors `BaseQueryService.getSearchQuery`; empty-means-unlimited is deliberate here, advertised by `placeholder="Unlimited"` across the selector modals |

Helias (approved):

